PR #17, opened 2017-03-08
Note editor Mobile view doesn't play for long enough

### LINKED ISSUE #16: Note editor Mobile view doesn't play for long enough.
Doesn't work well on IOS / Android.

chmora:
fixes #16 

The problem was that mouse events call preventDefault() and browser needs to wait for the events to finish. That's why now I use touch events to avoid that call and guarantee browser that the listener will not call preventDefault()

more information: 
www.chromestatus.com/features/5093566007214080
developers.google.com/web/updates/2017/01/scrolling-intervention

### DIFF
--- a/core/audio_context_manager.js
+++ b/core/audio_context_manager.js
@@ -38,11 +38,15 @@ var pxtblocky;
             return undefined;
         }
         function mute(mute) {
+            if (!_context)
+                return;
             _mute = mute;
             stop();
         }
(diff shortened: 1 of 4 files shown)


microsoftsam:
What's the deal with the warning message?

```
Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080
touchWrapFunc @ blockly.js:467
```

PR #20, opened 2017-03-22
Remove redundant listeners for category events

riknoll:
Was causing the workspace's resize function to be called ~50 times whenever a category was opened.


core/toolbox.js
```diff
@@ -520,6 +520,19 @@ Blockly.Toolbox.TreeControl.prototype.handleTouchEvent_ = function(e) {
  * @override
  */
 Blockly.Toolbox.TreeControl.prototype.createNode = function(opt_html) {
+    if (this.toolbox_ && !this.eventHandlerRegistered_) {
```
microsoftsam: Add a comment about this change, makes it easier when we have to merge with mainline blockly.

PR #33, opened 2017-05-23
Handle right click as a normal click

microsoftsam:


riknoll:
Is this just in the context menu?

microsoftsam:
Flyout and Toolbox. Not workspace.

PR #34, opened 2017-06-05
Toolbox can now be scrolled with touch gestures

guillaumejenkins:


microsoftsam:
Remove whitespace changes, and add comments wherever you diverge from blockly mainline.

microsoftsam:
Just thought of a scenario that isn't covered here. 
On a desktop if you mouse down and drag, it doesn't scroll on the toolbox. This works in the flyout, and in the workspace. Possibly something worth supporting as part of this change. @guillaumejenkins

core/toolbox.js
```diff
@@ -170,7 +170,8 @@ Blockly.Toolbox.prototype.init = function() {
           Blockly.hideChaff(true);
         }
         Blockly.Touch.clearTouchIdentifier();  // Don't block future drags.
-      });
+      }, /*opt_noPreventDefault*/ false, /*opt_noPreventDefault*/ true);
```
rachel-fenichel: I'm looking at pulling this into Blockly.  Why do you have the same argument twice with different values here?

microsoftsam: My mistake.

microsoftsam: @guillaumejenkins looks like there's a mistake here, can you take a look.

guillaumejenkins: It's a comment copy/paste mistake, the 1st parameter comment should read `/*opt_noCaptureIdentifier*/`, as seen in the function signature of `Blockly.bindEventWithChecks_` above

core/toolbox.js
```diff
@@ -537,9 +539,10 @@ Blockly.Toolbox.TreeControl.prototype.enterDocument = function() {
  * @private
  */
 Blockly.Toolbox.TreeControl.prototype.handleTouchEvent_ = function(e) {
-  e.preventDefault();
+  // pxtblockly: Remove preventDefaut() to allow scrolling toolbox with touch
```
rachel-fenichel: When you end a scroll, does it select the category that your pointer was on?  I think that would be fine for us--not perfect behaviour, but worth it to get scrolling on the toolbox.

What is the behaviour if you end the scroll by dragging off of the toolbox and then releasing?

Something feels wrong about calling `onMouseDown` as a result of a `TOUCHEND`, but I think it's reasonable.

core/toolbox.js
rachel-fenichel: Great, thanks for the clarification. 

I have a few other questions below to make sure I know exactly what I'm pulling in.

In this case you're setting opt_noPreventDefault because you want a click on the toolbox to dismiss any context menus and also potentially start a scroll, correct?

core/toolbox.js
guillaumejenkins: The following answers are for Chrome, IE & Edge; other browsers are untested
- Ending a scroll does not select a category
- Starting a scroll selects the category on Edge & IE
- Starting a toolbox drag and then moving the touch gesture outside the toolbox is correctly handled (as in, the toolbox correctly scrolls according to your touch gesture even if you dragged outside the toolbox)
- Ending a drag outside the toolbox does nothing in particular (flyout remains open if it was)
- It does indeed feel weird that a touch end calls on mouse down... it's been a while and I'm not sure why I changed this from START to END, but it seems like `handleTouchEvent_` is never called during a toolbox scroll for some strange reason, so it might be possible to revert that specific change
- The whole idea behind this commit was to let the browser handle the scroll (since the toolbox is a simple div), and to do that I had to remove the call to preventDefault. It does close any context menu as a side effect

Hope this helps clarify

core/toolbox.js
guillaumejenkins: Consolidated answers in the other comment

core/toolbox.js
rachel-fenichel: `handleTouchEvent_` isn't being called during a scroll because you also changed it to END above, where you're registering `handleTouchEvent_`.

PR #39, opened 2017-07-19
Re-enabling the mutator on if/else blocks

riknoll:
I think the goal of commenting out the mutator was to remove the cogwheel, but doing that breaks the logic block so that you can't define blocks with `else` or `elseif` in the toolbox. To get rid of the cogwheel, you just need to remove the `compose()` and `decompose()` functions which will leave the rest of the mutation intact.


riknoll:
Note that this is a PR for the "develop" branch

PR #42, opened 2017-08-04
Support zoom on pinch, using both pointer and touch events

samelhusseini:
First of all, browser support for touch gestures is crazy and all over the place.
- IOS support gesture apis, which returns the scale of the pinch through the event, and the browser handles all the complex maths.
- Android and most browsers on Surface / Surface books (except Edge) support touch events. 
- Edge support pointer events. (It seems most browsers are leaning towards adopting this standard but this hasn't quite made it)

All have their own glory and variations in the amount of flexibility they give the developer in tweaking the gesture.

I've implemented all three. Favouring the IOS gestures when on an IOS device. Using pointer events when that that is supported, and touch events in other cases.
Note: there are still some fine tunings that can be done, but this is a pretty good first attempt. 

@rachel-fenichel this might be something to consider bringing to mainline blockly.


rachel-fenichel:
@microsoftsam I'm definitely interested, and will take a look. 

In the past we've strongly avoided platform-specific code because increasing our number of code paths made testing difficult.  The move to Gestures may help because it's now a mapping onto the same underlying code.

Note that Apple has stated that Safari [will not support](https://lists.webkit.org/pipermail/webkit-dev/2012-December/023050.html) pointer events.  The pointer events polyfill exists to fill that gap but bloats the library.

rachel-fenichel:
From a quick glance at your code, I think it's in the wrong place for longer-term merge stability.  It should be in gesture.js, or (better) you should create a class that extends from gesture.js and has all of this code.

How does the new code interact with drags or pinches that start on blocks and fields?

I expect you've also got two listeners bound for touch start, move, and end, if you left the normal blockly listeners in.  In a touch environment Blockly binds events for both mouse and touch when you ask it to bind events for things like mouseDown.

microsoftsam:
Thanks for the quick feedback @rachel-fenichel, I'll look into splitting this out a little and moving it to a new file. 
As for your question about interaction with the current events, I ran into this issue with pointerevents as well. Both events are handled (hence why it's important to make sure I didn't call preventDefault), but the difference is the pinch gesture event is only relevant if there are two touches. So in some cases you could still zoom and drag at the same time, but I thought that was a fair side effect. I generally called cancel current gesture when I detected more than two touches.

microsoftsam:
The IOS gesturechange events don't interfere with any of the Blockly code which is nice.

pelikhan:
@microsoftsam status on this?

microsoftsam:
Need to do some refactoring, haven't had a change to look into this, will look into it in the next couple of weeks.

microsoftsam:
@rachel-fenichel this is ready. 

A summary of the changes made:
Since Blockly already splits all touch events into multiple events, this worked out well for converging the touchevent and pointer events path (as pointer events are handled one by one). 
I dropped support for the IOS specific gesturechange and gestureend events, touch events were enough for pinch to zoom support.

As advised, I've created a new touch_gesture file that extends gesture.js and added the support for pinch to zoom there. 

Here's the basic idea of how touch_gesture works: when a gesture is created, it adds a handler for another touchstart / pointerdown. This handler passes /*opt_noCaptureIdentifier*/ to ensure all events are handled (including those that don't meet the checkTouchIdentifier check. This handler is watching in order to convert a gesture into a "multi-touch" event. 
As with gesture.js it also adds handlers for move and end, but same as above, it passes noCaptureIdentifier to receive all events (needed for multi-touch).

I've made sure this doesn't conflict with dragging. If at any point a drag is going on, pinch to zoom is ignored. That is for both a workspace drag and a block drag.
Since the start, move and up events are handling all events (opt_noCaptureIdentifier), I've had to include Blockly.Touch.shouldHandleEvent in these event handlers to ignore multi-touch events in cases where we don't want to handle it (ie: when dragging a block or the workspace).

This PR also includes changes to support pointer events (I hadn't yet integrated the change from master to develop), the change solves it correctly. Handling checkTouchIdentifier to support pointer id.

I've also handled long hold properly here, with pointer event support. Fixes https://github.com/Microsoft/pxt-blockly/issues/90

When you get a chance, it'd be great if you could review this, and please ask me all the questions for anything that doesn't make sense.

This change passes eslint.

This change has been tested on: 
- [x] Surface Book (IE, Edge, Chrome, Firefox)
- [x] IOS 10 (iPad, iPhone)
- [x] Android (Chrome, Firefox for Android)
- [x] Mac (Chrome, Safari)

core/touch_gesture.js
```diff
@@ -0,0 +1,261 @@
...
+ * Sever all links from this object.
+ * @package
+ */
+Blockly.TouchGesture.prototype.dispose = function() {
+  Blockly.TouchGesture.superClass_.dispose.call(this);
+
+  if (this.onStartWrapper_) {
+    Blockly.unbindEvent_(this.onStartWrapper_);
+  }
+  if (this.onGestureChangeWrapper_) {
```
rachel-fenichel: What are `onGestureChangeWrapper` and `onGestureEndWrapper`?  I don't see them being set anywhere.

samelhusseini: Yep, dead code (was there when I had was handling gestureChange and gestureEnd for IOS)

core/touch_gesture.js
```diff
@@ -0,0 +1,261 @@
...
+ * Handle a touch start or pointer down event and keep track of current pointers.
+ * @param {!Event} e A touch start, or pointer down event.
+ * @package
+ */
+Blockly.TouchGesture.prototype.handleTouchStart = function(e) {
+  var pointerId = Blockly.Touch.getTouchIdentifierFromEvent(e);
+  // store the pointerId in the current list of pointers
+  this.cachedPoints[pointerId] = this.getTouchPoint(e);
+  var pointers = Object.keys(this.cachedPoints);
+  // If two pointers are down, check for pinch gestures
```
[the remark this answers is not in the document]
samelhusseini: Yes, if two pointers are down, we assume you are trying to zoom. In the future, if there are more multi-touch scenarios supporter this can be altered.

core/touch_gesture.js
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   * Boolean for whether or not this gesture is a multi-touch gesture.
+   * @type {boolean}
+   * @private
+   */
+  this.isMultiTouch_ = false;
+
+  /**
+   * A map of cached points used for tracking multi-touch gestures.
+   * @type {boolean}
```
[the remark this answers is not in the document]
samelhusseini: Yep, not the right type.

core/touch_gesture.js
```diff
@@ -0,0 +1,261 @@
...
+   * @private
+   */
+  this.cachedPoints = {};
+
+  /**
+   * A scale value tracking the previous gesture scale.
```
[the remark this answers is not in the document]
samelhusseini: I've seen it between 0 and 8.

samelhusseini:
Updated with PR feedback, found a few more style nits and tried to simplify the logic in handleMove.

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+ * Note: In this file "start" refers to touchstart, mousedown, and pointerstart
+ * events.  "End" refers to touchend, mouseup, and pointerend events.
+ */
+
+/**
+ * Class for one gesture.
+ * @param {!Event} e The event that kicked off this gesture.
+ * @param {!Blockly.WorkspaceSvg} creatorWorkspace The workspace that created
+ *     this gesture and has a reference to it.
+ * @constructor
```
rachel-fenichel: Missing annotation: `@extends {Blockly.Gesture}`

Within the annotation the whole class can disappear during compression.

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+  this.onMoveWrapper_ = Blockly.bindEventWithChecks_(
+    document, 'mousemove', null, this.handleMove.bind(this), /*opt_noCaptureIdentifier*/ true);
+  this.onUpWrapper_ = Blockly.bindEventWithChecks_(
+    document, 'mouseup', null, this.handleUp.bind(this), /*opt_noCaptureIdentifier*/ true);
+
+  e.preventDefault();
+};
+
+/**
+ * Handle a mouse down, touch move, or pointer move event.
```
rachel-fenichel: Update comment and param: handle a mouse down, touch start or pointer down event.

core/touch.js
```diff
@@ -194,7 +204,18 @@ Blockly.Touch.setClientFromTouch = function(e) {
  */
 Blockly.Touch.isMouseOrTouchEvent = function(e) {
   return goog.string.startsWith(e.type, 'touch') ||
-      goog.string.startsWith(e.type, 'mouse');
+      goog.string.startsWith(e.type, 'mouse') ||
+      goog.string.startsWith(e.type, 'pointer');
+};
+
+/**
+ * Check whether a given event is a touch event.
```
rachel-fenichel: is a touch event or pointer event.

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+    e.preventDefault();
+    e.stopPropagation();
+
+    this.dispose();
+  }
+};
+
+/**
+ * Whether this gesture is part of a mulit-touch gesture.
+ * @return {boolean} whether this gesture was a click on a workspace.
```
rachel-fenichel: Fix return description.  Nit on line above: "multi-", not "mulit-".

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+ * Helper function returning the current touch point coordinate.
+ * @param {!Event} e A touch or pointer event.
+ * @return {goog.math.Coordinate} the current touch point coordinate
+ * @package
+ */
+Blockly.TouchGesture.prototype.getTouchPoint = function(e) {
+  if (!this.startWorkspace_) {
+    return null;
+  }
+  var metrics = this.metrics_ || (this.metrics_ = this.startWorkspace_.getMetrics());
```
rachel-fenichel: Why do you need the metrics?  All of the work that you're doing with these pointer locations involves deltas, so you shouldn't need the workspace info.  There is on call to Blockly.utils.mouseToSvg (line 248), but that uses the event directly and not the stored coordinate.

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+    // Calculate the distance between the two pointers
+    var point0 = this.cachedPoints_[pointers[0]];
+    var point1 = this.cachedPoints_[pointers[1]];
+    var moveDistance = goog.math.Coordinate.distance(point0, point1);
+    var startDistance = this.startDistance_;
+    var scale = this.touchScale_ = moveDistance / startDistance;
+    
+    if (this.previousScale_ > 0 && this.previousScale_ < Infinity) {
+      var gestureScale = scale - this.previousScale_;
+      var delta = gestureScale > 0 ? gestureScale * 5 : gestureScale * 6;
```
rachel-fenichel: Can you explain this line?

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+  // If two pointers are down, check for pinch gestures
+  if (pointers.length == 2) {
+    // Calculate the distance between the two pointers
+    var point0 = this.cachedPoints_[pointers[0]];
+    var point1 = this.cachedPoints_[pointers[1]];
+    var moveDistance = goog.math.Coordinate.distance(point0, point1);
+    var startDistance = this.startDistance_;
+    var scale = this.touchScale_ = moveDistance / startDistance;
+    
+    if (this.previousScale_ > 0 && this.previousScale_ < Infinity) {
```
rachel-fenichel: This will always fail on the first call because previousScale was 0.  Intentional?

core/touch_gesture.js
```diff
@@ -0,0 +1,287 @@
...
+   * @private
+   */
+  this.cachedPoints_ = {};
+
+  /**
+   * A scale value tracking the previous gesture scale.
+   * @type {number}
+   * @private
+   */
+  this.previousScale_ = 0;
```
rachel-fenichel: This is the ratio between the starting distance between the touch points and the most recent distance between the touch points.  
Scales between 0 and 1 mean the most recent zoom was a zoom out.  
Scales above 1.0 mean the most recent zoom was a zoom in.

core/touch.js
```diff
@@ -74,14 +80,17 @@ Blockly.longPid_ = 0;
...
-  if (e.changedTouches.length != 1) {
+  if (e.changedTouches && e.changedTouches.length != 1) {
     return;
   }
   Blockly.longPid_ = setTimeout(function() {
-    e.button = 2;  // Simulate a right button click.
-    // e was a touch event.  It needs to pretend to be a mouse event.
-    e.clientX = e.changedTouches[0].clientX;
-    e.clientY = e.changedTouches[0].clientY;
+    if (e.changedTouches) {
```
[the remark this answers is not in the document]
samelhusseini: Yes and will do.

core/touch_gesture.js
samelhusseini: Good to know.

core/touch_gesture.js
samelhusseini: Ah nice catch.

core/touch_gesture.js
samelhusseini: Yes. We need the previous scale to figure out whether the intention is a zoom in or out.

core/touch_gesture.js
samelhusseini: Yeah this line is interesting, should probably be changed to use constants instead of inline values, but this is where the calibration happens. 

gestureScale gives us a scale but it's too small for a delta, so I use a multiplier to scale it so something that worked well (after doing some calibration on a bunch of touch devices). I'm scaling it differently whether it's a zoom in or a zoom out.

I think for starters, the multipliers should be moved to constants, and we can continue to calibrate these values to something that works well. These haven't been perfected yet.

core/touch_gesture.js
samelhusseini: I think you might be right, since the touch point is all relative, the distance can be calculated irrespective of the workspace's absolute position. I think I needed the metrics at some point for the zoom, but as you mentioned I'm  using mouseToSvg above instead.

core/touch_gesture.js
```diff
@@ -0,0 +1,297 @@
...
+   * at the end of a drag. Opaque data returned from Blockly.bindEventWithChecks_.
+   * @type {Array.<!Array>}
+   * @private
+   */
+  this.onStartWrapper_ = null;
+};
+goog.inherits(Blockly.TouchGesture, Blockly.Gesture);
+
+/**
+ * A multiplier used to convert the gesture scale to a zoom in delta
```
rachel-fenichel: Nit: end comment with period, here and for the ZOOM_OUT_MULTIPLIER.

samelhusseini:
Yay! Thanks for reviewing!

PR #48, opened 2017-08-14
Fix wrong width on dropdowns with an image selection

### LINKED ISSUE #46: BorderRect broken when an icon is selected in a dropdown (Edge & IE)
Taken from https://github.com/google/blockly/issues/1274 and re-posted here for tracking

Expected:
image

Actual:
image

[Line 345 of block_render_svg.js](https://github.com/google/blockly/blob/master/core/block_render_svg.js#L345) forces an extra width update on IE & Edge for fields. The problem is, the forced update attempts to get the cached width, which ends up in [the catch block in field.js](https://github.com/google/blockly/blob/master/core/field.js#L361) because of the exception that `getComputedTextLength()` throws on IE / Edge. The width is then approximated only based on the text content (` ▾`), without considering the width of the selected image.

Possible fixes would be to not force the width update if the field has non-text content, or make the width approximation in `getCachedWidth()` take into account non-text content in fields.

guillaumejenkins:
Fixes #46

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -420,3 +420,24 @@ Blockly.FieldDropdown.prototype.dispose = function() {
   Blockly.WidgetDiv.hideIfOwner(this);
   Blockly.FieldDropdown.superClass_.dispose.call(this);
 };
+
+/**
+ * pxt-blockly: Added this method to fix image selection bug on IE / Edge
+ * Updates the width of the field. Overrides field.prototype.updateWidth to
+ * deal with image selections on IE and Edge. If the selected item is not an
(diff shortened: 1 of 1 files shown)


microsoftsam:
Looks good. @rachel-fenichel RE https://github.com/google/blockly/issues/1274

rachel-fenichel:
@guillaumejenkins please make a pull request with this code against blockly's develop branch.

Three nits:
- The end of the block quote documenting the function should have one `*`, not two.
- Remove the "pxt-blockly" comment for main blockly.
- Why are you only updating the width and not the height?

PR #53, opened 2017-09-12
cleaning up the math

pelikhan:
Sanitizing the math. Waiting for closure to allow building again.


pelikhan:
I can't build. This is annoying.

PR #61, opened 2017-09-25
Highlight only a single block rather than the entire stack.

samelhusseini:
Set the highlight on the block svg rather than the entire stack


microsoftsam:
see linked PXT PR for screenshots: https://github.com/Microsoft/pxt/pull/3031

guillaumejenkins (approved):

PR #64, opened 2017-09-28
Turn cursor into a pointer if there's a click handler on the field image

samelhusseini:
Hard to take a screenshot but title says it all.


guillaumejenkins (approved):

microsoftsam:
@rachel-fenichel simple fix, probably makes sense to be in mainline.

PR #66, opened 2017-09-30
Dropdown color picker

microsoftsam:
Change the Field_color to use the scratch-blocks dropdown instead. 
Altering CSS to make cells a little bigger.

screen shot 2017-09-30 at 6 42 20 pm


pelikhan:
Can we reduce to 5x5? Bottom row is clipped

microsoftsam:
Thats because theres more rows and a scrollbar. Theres too many color choices IMO i will try with fewer cooors

PR #71, opened 2017-10-13
If statement add button to add else first

microsoftsam:
If statement add button to first add an else block and then continue to add else-ifs.

addbtnelse

Fixes https://github.com/Microsoft/pxt/issues/3220


pelikhan:
LGTM!

PR #73, opened 2017-10-23
Don't allow empty procedure names

### LINKED ISSUE #72: Function name field can be edited to be blank
The function block allows users to edit it to be blank.

PXT issue: https://github.com/Microsoft/pxt/issues/3265

microsoftsam:
If a user edits a procedure name to be blank, we replace it with the original name of the procedure. 

Fixes #72


microsoftsam:
https://github.com/Microsoft/pxt/issues/3265

PR #74, opened 2017-10-23
Better highlighting

microsoftsam:
Better highlight solved by copying the SVG path of a block above the entire stack, with a transparent background, and applying a filter.

betterhighlight


microsoftsam:
@brachbach

Issue #84, opened 2017-11-08
nested get_variable block breaks a number of things in IE


microsoftsam:
There's a bizarre issue with blocks that have get_variable blocks inside of them in IE. 
An example of this is the text_indexOf block in the playground. 
It's definition is: 
```
      <block type="text_indexOf">
        <value name="VALUE">
          <block type="variables_get">
            <field name="VAR">text</field>
          </block>
        </value>
        <value name="FIND">
          <shadow type="text">
            <field name="TEXT">abc</field>
          </shadow>
        </value>
      </block>
```

In IE, when this block is in the toolbox. A few things break, scrolling in the flyout no longer works, and when you drag, the filter applied on the block makes it hidden: 

![ievariablefields](https://user-images.githubusercontent.com/16690124/32565237-4d111b0e-c46b-11e7-8a5d-060e7c5121d2.gif)

Simply dragging out the get_variable block and re-connecting it fixes the issue of the disappearing block. 

Another workaround is to make the get_variable block a shadow block, although Blockly has an assert to avoid this (not sure why), but if we remove the assert scrolling and dragging work fine. 

Note: another block that suffers from similar behaviour. is the if / else block. Perhaps the same root cause.

@rachel-fenichel I'm pretty clueless as to what might be going on, any pointers?
I imagine this would repro in scratch-blocks, but I don't think scratch is using any nested get_variable blocks inside of the flyout.

rachel-fenichel:
Is #83 a dupe of this?

This looks like a serious bug and I've got a couple of guesses, but I don't have a specific answer off the top of my head.

First off, I would expect it to repro in scratch-blocks, but also in Blockly on the playground.  Blockly and Scratch-blocks should be doing very similar things here.  Can you check if it happens in Blockly?

The if/else block doesn't have any variables, right?  Is it the same as the one on scratch-blocks?

If it were just variables I'd say we had a problem with the code that creates a variable as you drag in.  If it were just dragging blocks in IE I'd say the drag surface had a problem.  But not being able to scroll doesn't fit with either of those hypotheses.

Making get_variable a shadow block breaks in this case:
- drag out a block with a shadow variable block (creating the variable)
- cover the shadow
- delete the variable
- uncover the shadow

There's a forum thread about it that goes into more detail, but we're pretty convinced that the restriction is the right choice.

microsoftsam:
Yes #83 was a dupe. Github is playing up on me today. 
Agreed, that it looks serious.

[...]

PR #85, opened 2017-11-08
Fix IE rendering issue causing scrollbar and drag issues

samelhusseini:
Fix issue in IE where the scrollbar and drag breaks in certain cases where an empty input path exists in the rendered block.
As described in https://github.com/Microsoft/pxt-blockly/issues/84, the cases that this appears in our codebase were the ifelse statement and any block that had a get_variable in it. 

The fix here is to define an empty data attribute for the path on initialisation of the path. According to https://www.w3.org/TR/SVGTiny12/paths.html this means the path won't be rendered by the browser, which is okay in this case since we're setting the visibility to hidden anyway. 

Tested on: 
- [x] Edge 16
- [x] Chrome Mac
- [x] Chrome Windows
- [x] Safari
- [x] Firefox Mac
- [x] Firefox Windows

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -153,7 +153,8 @@ Blockly.BlockSvg.prototype.initSvg = function() {
     for (var i = 0, input; input = this.inputList[i]; i++) {
       input.init();
       if (input.type === Blockly.INPUT_VALUE) {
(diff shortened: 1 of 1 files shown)


microsoftsam:
Wrong PR.

PR #86, opened 2017-11-08
Fix IE rendering issue causing scrollbar and drag issues

microsoftsam:
Fix issue in IE where the scrollbar and drag breaks in certain cases where an empty input path exists in the rendered block.
As described in https://github.com/Microsoft/pxt-blockly/issues/84, the cases that this appears in our codebase were the ifelse statement and any block that had a get_variable in it. 

The fix here is to define an empty data attribute for the path on initialisation of the path. According to https://www.w3.org/TR/SVGTiny12/paths.html this means the path won't be rendered by the browser, which is okay in this case since we're setting the visibility to hidden anyway. 

Tested on: 
- [x] Edge 16
- [x] Chrome Mac
- [x] Chrome Windows
- [x] Safari
- [x] Firefox Mac
- [x] Firefox Windows
- [x] and obviously IE (11)

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -189,7 +189,8 @@ Blockly.BlockSvg.prototype.initInputShape = function(input) {
     'path',
     {
       'class': 'blocklyPath',
-      'style': 'visibility: hidden' // Hide by default - shown when not connected.
+      'style': 'visibility: hidden', // Hide by default - shown when not connected.
+      'd': '' // IE doesn't like paths without the data definition, set an empty default
     },
     this.svgGroup_
   );


microsoftsam:
@rachel-fenichel

rachel-fenichel:
LG, and this should go into scratch-blocks, but I'm still confused about why this would be restricted to variable getters.  Based on what you've said I would expect the problem to occur when loading from XML in IE for any blocks that have a real (non-shadow) block connected to a value input.  

For instance, I would expect loading this to fail in IE on the [scratch-blocks playground](http://llk.github.io/scratch-blocks/playgrounds/tests/vertical_playground.html):

```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <variables></variables>
  <block type="operator_round" id="O+_OKjU#VqyT;$UQEtpj" x="213" y="663">
    <value name="NUM">
      <shadow type="math_number" id="jYR3GIY=+[sf4iU!yY^f">
        <field name="NUM"></field>
      </shadow>
      <block type="operator_round" id="n99XF_w-1PyTB-G^Li#B">
        <value name="NUM">
          <shadow type="math_number" id="2XKLqR:f/`|rFbaV!`i`">
            <field name="NUM"></field>
          </shadow>
        </value>
      </block>
    </value>
  </block>
</xml>
```

microsoftsam:
I'm not sure why it's restricted to variables either, but from what I've gather in the case of variable inputs, the rendering logic creates the DOM element for the input, but doesn't render it (as there's a block to be rendered inside of it, hence it doesn't create the path for it). 
But if you disconnect it (it renders it), and re-connect it, the path data stays there, hidden behind the rendered child block. Hence why it doesn't repro after a re-render.

The above XML doesn't repro in scratch-blocks.
It's possible the above fix is just **masking** the issue, and there really shouldn't be an extra input field DOM in the first place.

Here's some valid user XML that does repro in scratch-blocks: 
```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <variables>
    <variable type="" id="k4F(%ICqJI8{TMJTT^2$">i</variable>
  </variables>
  <block type="looks_switchbackdropto" id="1EI0u[`mJo[63tjYW~1^" x="212" y="147">
    <value name="BACKDROP">
      <block type="data_variable" id="=G^yxBY/_D/krG%Cr(xa">
        <field name="VARIABLE" id="k4F(%ICqJI8{TMJTT^2$" variabletype="">i</field>
      </block>
    </value>
  </block>
</xml>
```

microsoftsam:
Actually, the interesting thing about that scratch-blocks repro, and also the repro that I'm seeing in PXT is that there is **no** shadow block behind the field variable.

Which makes this repro much less likely in scratch-blocks as all inputs have shadow blocks.

screen shot 2017-11-08 at 3 35 11 pm

rachel-fenichel:
The "input shape" referenced in code is that darker oval (or hexagon on other blocks).  I thought it was there on all blocks, regardless of whether they had a shadow.

One more repro check: what about this one?

```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <variables></variables>
  <block type="operator_and" id="{o/zPw8mbvA4/K#l:DY#" x="219" y="340">
    <value name="OPERAND2">
      <block type="operator_not" id=";;WvgQM]fo%[WIC=JHj$"></block>
    </value>
  </block>
</xml>
```
image


Booleans don't have shadow blocks, unlike string and number inputs.  If that fails and the previous one I posted succeeds, it's definitely about input shapes when there is a block but not a shadow block.

microsoftsam:
Yep that failed.

microsoftsam:
So do you think there's an underlying issue? or do we just work around the IE issue by ensure the d attribute is defined.

rachel-fenichel:
I think there may be an underlying issue, but that you should work around the IE issue by merging this fix.  

I'll file a bug on scratch blocks with those two cases, and I can now debug without firing up my IE box, since I know the problem comes from the missing `d` attribute on the `inputShape` path.  I've confirmed that I can see that in chrome--it's just that it doesn't choke on it.

Thanks for all the help!

microsoftsam:
No worries. Thanks for your help as well :)

rachel-fenichel:
Do you know why filter:none fixed it/hid the problem, by the way?

microsoftsam:
I don't really, but this isn't the first time I've seen mid-drag disappearing issues. 
A couple months back, I saw this issue quite consistently in an Electron app setting (with Chromium 56)
I hadn't managed to get as far as I did today, but setting filter:none to the svg.blocklyBlockDragSurface > g avoided the issue altogether. ie: the disappearance only appeared as the filter was applied.

It's possible that a path without the data messes with the filter operations of the browser. I'm not 100% sure.

microsoftsam:
btw, how did you get the Malformed SVG: missing `d` attribute on a `path error in Chrome? What flag do I need to tick to get these errors to show up?

rachel-fenichel:
I made up the phrasing, but I'm glad it sounds like a real error message :)

I loaded the two cases above and used the developer tools to confirm that `d` was not set in Chrome.  I had the hunch that if it was missing in IE, it was probably also missing in Chrome, but that Chrome was recovering.

microsoftsam:
Haha nice!

PR #92, opened 2017-11-18
Support pointer events

microsoftsam:
Now that Chrome 60+ supports pointer events, it's a good time to support this in Blockly.

Import note when supporting pointer events, is to make sure all elements have touch-action:none otherwise the events get cancelled by the device.

Good article RE pointer events: https://developers.google.com/web/updates/2016/10/pointer-events

@rachel-fenichel 

Tested on: 
- Android with Chrome 62
- Edge 16 on Surface Book 
- Chrome 62 on Surface Book 
- Chrome on Mac
- IOS 10
- Safari


rachel-fenichel:
I'm making a PR on blockly with these changes right now, and it looks like some things are missing.

For instance, the touch identifier is a core part of how the gesture code works, but there's no code here to use a pointer event's `pointerId', which is the equivalent field.

It looks like this works because the touch identifier check is never happening.  `Blockly.Touch.shouldHandleEvent` always returns true if the event is not a mouse or touch event, because we always want to handle key events, etc.

That means that `checkTouchIdentifier` is never called, so `getTouchIdentifierFromEvent` is never called.  That's where I'd expect the `pointerId` reference to be.

I haven't verified on a device with pointer events, but I expect that they won't handle multiple touches well right now.  I would expect them to not throw out secondary touches, so putting a second finger down during a drag would make the dragged block bounce back and forth between the two active pointers.  Is that true?  If not, what code is making it work?

microsoftsam:
Thanks @rachel-fenichel, will look into this.

PR #95, opened 2018-01-19
Fix rendering and positioning of hidden inputs

riknoll:
Fixes two bugs:

1. When a value input with no connected block is hidden, the input shape (the thing that is drawn in place of a block) stays rendered while the rest is hidden.
2. When a value input with a connected block is hidden and the parent block is moved, the connection tightening logic was translating the block's SVG but not moving the block's connections. That ended up causing a feedback loop where a block would move way out of position and render incorrectly once the input was un-hidden.

@rachel-fenichel I don't know if these impact mainline Blockly or not; is there a testing process that Sam usually goes through to determine that?

### DIFF
--- a/core/input.js
+++ b/core/input.js
@@ -182,6 +182,13 @@ Blockly.Input.prototype.setVisible = function(visible) {
   }
   this.visible_ = visible;
 
+  // pxtblockly: hidden inputs get filtered out of the block's
+  // render list so the input shapes don't get hidden. Hide them
+  // here instead
+  if (!visible && this.outlinePath) {
+    this.outlinePath.setAttribute('style', 'visibility: hidden');
+  }
+
   var display = visible ? 'block' : 'none';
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
To check if the same problem exists in Blockly or Scratch-Blocks you can use the playgrounds.  The [Blockly](https://blockly-demo.appspot.com/static/tests/playground.html) one is updated whenever we do a release.  The [scratch-blocks](http://llk.github.io/scratch-blocks/playgrounds/tests/vertical_playground.html) one updates more regularly.

In both cases you can use `Blockly.getMainWorkspace()` from the console to get the workspace, then directly get blocks and make calls on them if you need to do something programmatically.

For this particular problem: It looks like it would definitely be a problem in scratch-blocks, which has input shapes.  I'm not sure if it's a problem in Blockly, but I don't think so.

PR #110, opened 2018-08-07
Choose connection targets based on cursor location during drag 

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/910

We have received a lot of feedback that for large blocks it's difficult to figure out what part of the block needs to overlap an input to get it to connect. In vanilla Blockly it was pretty unambiguous because you just had to line up the puzzle pieces but in Scratch blocks the left side looks just like the right.

This PR is an attempt at making that behavior feel a little more natural by basing the connection logic on the location where the user grabs the block and not the location of the output connection. 

Here is the old behavior:

drag_offset_old

Here is the new behavior:

drag_offset_new

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -39,9 +39,11 @@ goog.require('goog.asserts');
  * are being dragged by a mouse or touch.
  * @param {!Blockly.BlockSvg} block The block to drag.
  * @param {!Blockly.WorkspaceSvg} workspace The workspace to drag on.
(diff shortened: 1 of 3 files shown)


guillaumejenkins (approved):

jaustin:
I think this looks like a really good improvement and definitely worth deploying to test.

The only other solutions I can think of are more radical and quite probably not easier - for example when you click on the block to drag it we could shrink it down to the small size until dropped :/ ?

dragonx:
Yasssssssss!!!

PR #112, opened 2018-08-16
Bring back mutation to text_join

guillaumejenkins:
(Ignore the fact I changed the text on the block, that's another discussion):

joinmutation

I based these changes on the existing create_array_with logic

See https://github.com/Microsoft/pxt-microbit/issues/996

### DIFF
--- a/blocks/text.js
+++ b/blocks/text.js
@@ -29,6 +29,7 @@ goog.provide('Blockly.Constants.Text');
 goog.require('Blockly.Blocks');
 goog.require('Blockly');
 
+goog.require('Blockly.PXTBlockly.Extensions');
 
 /**
  * Unused constant for the common HSV hue for all blocks in this category.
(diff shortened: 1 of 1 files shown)


blocks/text.js
```diff
@@ -749,35 +750,138 @@ Blockly.Constants.Text.TEXT_JOIN_MUTATOR_MIXIN = {
...
+      if (firstInput && firstInput.connection.targetConnection) {
+        // Create a new shadow DOM with the same type as the first input
+        // but with an empty default value
+        var newInput = this.getInput('ADD' + (this.itemCount_ - 1));
+        var shadowInputDom = firstInput.connection.getShadowDom();
+        if (shadowInputDom) {
+          var shadowDom = document.createElement('shadow');
+          var shadowInputType = shadowInputDom.getAttribute('type');
+          shadowDom.setAttribute('type', shadowInputType);
+          var shadowDomField = document.createElement('field');
```
riknoll: Not sure why this field is being added. Looks like the field for a math_number block but the block type is coming from the shadow dom. Also, I think creating the field and not setting the content doesn't do anything

riknoll (changes requested):

riknoll (approved):

PR #114, opened 2018-09-05
Adding the abillity to remove the "else" in an if/else block

### LINKED ISSUE #1037: The "else" part of the "if-else" block should be removable
**Describe the bug**
In v1, after an "if" block is created, if the user accidentally adds an "else" part vis the "+" button, effectively making the block to an "if-else" block, there is no "-" button to remove the "else" part.

**To Reproduce**
1. Create an "if" block
2. Press the "+" button to add an "else" part

**Expected behavior**
There is a "-" button to remove the "else" part.

**Screenshots**
if-block

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1037

if_else_remove

Also fixing a bug with the RIGHT_ALIGN flag for inputs which I don't think we actually use anywhere else.

### DIFF
--- a/blocks/logic.js
+++ b/blocks/logic.js
@@ -128,6 +128,14 @@ Blockly.Blocks['controls_if'] = {
     this.update_(update);
     this.restoreConnections_();
   },
+  removeElse_: function() {
+    this.storeConnections_();
+    var update = function() {
+      this.elseCount_--;
+    };
+    this.update_(update);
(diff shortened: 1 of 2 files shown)


FranklinWhale:
With this change, I think the "if-else" block on the block selection panel may be removed as it can now be replaced by the "if" block.

guillaumejenkins (approved):

PR #119, opened 2018-09-25
Fix issue with removing the else clause when there was a block connected in that input

samelhusseini:
In an if-else statement, when a user tried to remove an else clause and there was a block connected to the if block, it would cause an exception because we were trying to reconnect a block to an input that no longer existed. 

This checks for that input before attempting to reconnect the else clause.

Fixes https://github.com/Microsoft/pxt-ev3/issues/804


pelikhan:
Nice bug!

PR #126, opened 2018-10-11
Update widget div positioning logic to not cut off content above the screen

samelhusseini:
The current positioning of the widget div would rather show content close to the top of the field and have it cut off the top of the screen, than show it above the entire field. 

Changing this logic to show content above a field in that case. 
Here's an example: 

screen shot 2018-10-10 at 6 03 36 pm


riknoll (approved):

pelikhan:
Does this also apply to the sprite editor positioning?

samelhusseini:
No, this is really only useful for the gridpicker which uses the WidgetDiv to place its content. 
Most new pickers use the DropdownDiv which adds an arrow that points to what field opened it. In that case showing it above the field might look weird, but we could also possibly just not show an arrow in that scenario. 

Anyway short answer, different code paths, this only fixes the WidgetDiv positioning.

PR #128, opened 2018-10-11
Darken shadow field input when the menu is open

samelhusseini:
In the same way that we darken menu dropdowns when a field is on a block. Do it when it's a shadow block. 

shadowdropdown


riknoll (approved):

pelikhan:
Yes

PR #137, opened 2018-10-15
Align elseif clause minus button to the right.

samelhusseini:
Avoid 

screen shot 2018-10-15 at 9 09 37 am

Instead
screen shot 2018-10-15 at 9 10 43 am


pelikhan (approved):
Good fix.

PR #143, opened 2018-10-15
Transition animation when centering on a block

samelhusseini:
Add a transition animation when moving to centre on a block in the workspace. 
transition

Note: this doesn't work in IE and Edge as they don't support SVG CSS transition on transforms.


pelikhan (approved):
Nice!

PR #144, opened 2018-10-17
Add hover effect to the draggable variable

samelhusseini:
Add hover effect to the draggable variable that shows a dotted outline around the variable when hovered. 
It also changes the cursor to the 'copy' cursor when hovered, see gif: 

draggablehover

### DIFF
--- a/core/gesture.js
+++ b/core/gesture.js
@@ -1004,6 +1004,7 @@ Blockly.Gesture.prototype.duplicateOnDrag_ = function() {
       newVariableField.textContent = xmlBlockField.textContent;
       newVariableBlock.appendChild(newVariableField);
       xmlBlock = newVariableBlock;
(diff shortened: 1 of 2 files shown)


core/gesture.js
```diff
@@ -1004,6 +1004,7 @@ Blockly.Gesture.prototype.duplicateOnDrag_ = function() {
       newVariableField.textContent = xmlBlockField.textContent;
       newVariableBlock.appendChild(newVariableField);
       xmlBlock = newVariableBlock;
+      this.targetBlock_.inputList[0].fieldRow[0].clearHover();
```
riknoll, line 1007: No guards needed here?

samelhusseini: This code path is only for that one block. Lots of assumptions here.

riknoll (approved):

PR #147, opened 2018-10-23
Always position dropdown wherever you have the most space

riknoll:
If the dropdown div can't fit below the block, Blockly puts it above the block instead. If it also can't fit the div above the block, it gives up and puts the div below.

This change tweaks that logic so that when there isn't enough space above and below, we choose whichever side has the most available space (instead of always below). The div still gets cut off, but the old behavior would often look pretty broken:

image

Now that looks more like this:

image

Definitely not great but maybe better? @samelhusseini I will default to your judgement over whether or not to merge this as a holdover fix.  Ultimately we should consider a different experience for small screens; maybe a modal centered in the workspace

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -275,9 +275,11 @@ Blockly.DropDownDiv.getPositionMetrics = function(primaryX, primaryY, secondaryX
   // renderX, renderY will eventually be the final rendered position of the box.
   var renderX, renderY, renderedSecondary;
   // Can the div fit inside the bounds if we render below the primary point?
-  if (primaryY + divSize.height > boundPosition.top + boundSize.height) {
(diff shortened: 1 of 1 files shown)


samelhusseini:
I made a change recently that fixes this for the positioning of widget divs. I've also discussed this with WW, and my thoughts are that the dropdown div should show on top of the page (ie: hiding the field) where you can't place it above or below. 

I think this would be a better fallback option:
https://github.com/Microsoft/pxt-blockly/pull/126

samelhusseini:
The difference between the dropdowndiv and the widgetdiv is that the dropdown div adds an arrow, and I think in the fallback case where it shows it up on top of the field, we should hide the arrow.

samelhusseini:
How about this change instead?
https://github.com/Microsoft/pxt-blockly/pull/148

Issue #57, opened 2017-09-15
Non-shadow string fields expand when clicked


guillaumejenkins:
![string expand bug](https://user-images.githubusercontent.com/14299377/30495863-8287a458-9a02-11e7-93c2-4a23de36a768.gif)

PR #150, opened 2018-10-23
Fix placement of text inputs for text fields on a block

### LINKED ISSUE #1457: Editing function name takes up the whole size of the block
**Describe the bug**
Function name editing takes up the entire block size. 

screen shot 2018-10-17 at 11 17 09 pm

**To Reproduce**
1. Make a function
2. Add a few blocks in the function (or not, still repro's)
3. Try to edit the function name
4. See that the input box takes up the entire block

**Expected behavior**
Only take up size of the field

Note: this bug affects all targets

samelhusseini:
We previously had issues with centering and a text field when it was a single field block. https://github.com/Microsoft/pxt-blockly/pull/108 attempted to fix this by centering a field using the original blocks height and width. This worked fine for single field blocks but broke text fields on blocks that had more than a single field (eg: the function block). 

This PR fixes that by aligning by only applying the "center in the middle of the block" logic to single field blocks.

Fixes https://github.com/Microsoft/pxt-blockly/issues/57
Fixes https://github.com/Microsoft/pxt-microbit/issues/1457


JoshuaLowe1002:
Works perfectly in EduBlocks. Thanks Sam! 
image

riknoll (approved):

PR #154, opened 2018-10-26
Perf improvements

samelhusseini:
Apply a number of performance improvements from Blockly. 
https://github.com/google/blockly/pull/1844/files
https://github.com/google/blockly/pull/1875/files

Also cache the size of icons in Flyout button. 
Also cache the field click target. 
@rachel-fenichel do you see any issue in caching the field's click target? Can it change in the field's lifecycle?

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -287,11 +287,11 @@ Blockly.Field.prototype.updateEditable = function() {
   if (this.sourceBlock_.isEditable()) {
     Blockly.utils.addClass(group, 'blocklyEditableText');
     Blockly.utils.removeClass(group, 'blocklyNonEditableText');
-    this.fieldGroup_.style.cursor = this.CURSOR;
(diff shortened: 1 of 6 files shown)


core/field.js
```diff
@@ -768,25 +766,21 @@ Blockly.Field.prototype.setTooltip = function(_newTip) {
  * @private
  */
 Blockly.Field.prototype.getClickTarget_ = function() {
+  if (this.clickTarget_) return this.clickTarget_;
```
samelhusseini, line 769: @rachel-fenichel here.

rachel-fenichel: Not sure--I feel like this can vary a lot by field.  But I think letting it be the svg root here is fine in the general case, and if it comes up for other fields they can override.

samelhusseini: Makes sense.

riknoll (approved):
LGTM, but I don't know much about these code paths

PR #158, opened 2018-10-26
Fix editing text fields on Android.

samelhusseini:
Keyboard makes us lose focus making editing impossible.

PXT-blockly PR of https://github.com/LLK/scratch-blocks/pull/1743

Tested on Android 7.1.1


riknoll (approved):
Well that was fast

PR #163, opened 2018-11-02
Help Flyout button

samelhusseini:
Support adding a help button in the flyout label. 
Here's what it looks like: 

screen shot 2018-11-02 at 11 24 15 am

Clicking on the question mark image (and only that) will trigger the button callback

Also added some more examples of using labels with lines / icons in the playground.


samelhusseini:
Ping for a review.

pelikhan (approved):
Why this feature?

samelhusseini:
For WW. https://github.com/playi/pxt-wonder/issues/337

PR #165, opened 2018-11-05
scrollOnCenter for large blocks

pelikhan:
Invalid use of getHeightWidth (super confusing name btw) lead to incorrect position calculation.


samelhusseini (approved):

samelhusseini:
A bit more context: 
the current centering logic breaks down if a block is larger than the screen height
this fix is not ideal, but we'll dig up a better fix in the near future.

samelhusseini:
https://github.com/Microsoft/pxt/pull/4931

PR #169, opened 2018-11-13
Reposition the dropdown div on a resize to fix issues with Android

samelhusseini:
https://github.com/Microsoft/pxt-blockly/pull/158 recently was a similar fix that repositioned the WidgetDiv when a resize occurred rather than closing it, which fixed issues with editing text fields on Android when the keyboard popped up. 

This PR adds a similar fix but for the Dropdown div so you're able to use the textdropdown field and other similar fields that focus on a text field which in turn pops up the keyboard. 

I'm also trying to standardize the ways we show the dropdowndiv into two methods: 
- showPositionedByBlock (Already exists) which positions the dropdown div around the block bounds
- showPositionedByField (New) which positions the dropdown div around the field. Which is currently only used by field_dropdown and field_number's numPad.

Field's should no longer directly call the dropdown.show method, and we should consider making it private at a later stage. 

@rachel-fenichel @picklesrus FYI.

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -183,13 +183,13 @@ Blockly.DropDownDiv.setCategory = function(category) {
  * by a particular block. The primary position will be below the block,
  * and the secondary position above the block. Drop-down will be
  * constrained to the block's workspace.
- * @param {Object} owner The object showing the drop-down
+ * @param {!Blockly.Field} field The field showing the drop-down
  * @param {!Blockly.Block} block Block to position the drop-down around.
  * @param {Function=} opt_onHide Optional callback for when the drop-down is hidden.
  * @param {Number} opt_secondaryYOffset Optional Y offset for above-block positioning.
  * @return {boolean} True if the menu rendered below block; false if above.
  */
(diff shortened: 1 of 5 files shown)


core/dropdowndiv.js
```diff
@@ -200,9 +207,43 @@ Blockly.DropDownDiv.showPositionedByBlock = function(owner, block,
...
+  return Blockly.DropDownDiv.show(owner, primaryX, primaryY, secondaryX, secondaryY, opt_onHide);
 };
 
+
+/**
+ * Shortcut to show and place the drop-down with positioning determined
+ * by a particular field. The primary position will be below the field,
+ * and the secondary position above the field. Drop-down will be
+ * constrained to the block's workspace.
+ * @param {Object} owner The object showing the drop-down
```
riknoll, line 219: Shouldn't this be field?

riknoll (approved):

core/dropdowndiv.js
samelhusseini: Changing to field for clarity.

PR #173, opened 2018-11-13
Don't show the numpad if the slider is shown

samelhusseini:
The slider field calls into the number field and in the case where we're on a touch device, it shows the num pad but then immediately replaces the content with a slider. 
Here: https://github.com/Microsoft/pxt-blockly/blob/develop/core/field_slider.js#L105

Skipping that step as it's unnecessary.

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -163,7 +163,7 @@ Blockly.FieldNumber.prototype.setConstraints_ = function(opt_min, opt_max,
 Blockly.FieldNumber.prototype.showEditor_ = function(e, opt_showNumPad) {
   Blockly.FieldNumber.activeField_ = this;
   // Do not focus on mobile devices so we can show the num-pad
-  var showNumPad = opt_showNumPad ||
+  var showNumPad = (typeof opt_showNumPad !== "undefined") ? opt_showNumPad :
       (goog.userAgent.MOBILE || goog.userAgent.ANDROID || goog.userAgent.IPAD);
   Blockly.FieldNumber.superClass_.showEditor_.call(this, e, false, showNumPad);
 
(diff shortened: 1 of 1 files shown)


core/field_number.js
```diff
@@ -157,13 +157,14 @@ Blockly.FieldNumber.prototype.setConstraints_ = function(opt_min, opt_max,
...
+ * @param {boolean=} opt_showNumPad If true, show the num pad.
  * @private
  */
-Blockly.FieldNumber.prototype.showEditor_ = function(e) {
+Blockly.FieldNumber.prototype.showEditor_ = function(e, opt_showNumPad) {
   Blockly.FieldNumber.activeField_ = this;
   // Do not focus on mobile devices so we can show the num-pad
-  var showNumPad =
-      goog.userAgent.MOBILE || goog.userAgent.ANDROID || goog.userAgent.IPAD;
+  var showNumPad = opt_showNumPad ||
```
guillaumejenkins: This change allows you to force the numpad to show, but not force it to **not** show (i.e. if you set the flag to false, it will still be shown on mobile)

If you want both force hide and force show, you'll need 2 separate flags

If you just want the ability to force hide, then maybe change the flag to `skipNumPad` and `and` it with the mobile checks?

guillaumejenkins (changes requested):
If I understand correctly, this new logic does not seem to accomplish what you want

core/field_number.js
samelhusseini: Too true.. I'll take a look. Thanks!

samelhusseini:
Updated

guillaumejenkins (approved):

PR #175, opened 2018-11-20
Fix clearing the workspace if an edit is being made

samelhusseini:
A recent change to fix simulator re-runs introduced a bug where we're sending a UI event when a block is disposing. This doesn't affect us today as we don't clear the workspace programmatically, but it affects @guillaumejenkins's work.


guillaumejenkins (approved):

samelhusseini:
I'm going to pull this in, but as it doesn't affect pxt today. I won't integrate today. It's safe to be integrated the next time we pull pxt-blockly in.

PR #177, opened 2018-12-05
Add an option to create a variable from the field_variable dropdown

samelhusseini:
Also adding a separator to group the variables, and the commands separately. 

screen shot 2018-12-05 at 9 18 10 am

There also seems to be an unnecessary margin to the right of all dropdown fields. The CSS is copied from closure and closure has the concept of a "shortcut" for each MenuItem which isn't used in Blockly. That margin is okay for dropdown's with smaller text. ie: the operator dropdown for the Math operator blocks, but it makes the field_variable dropdown unnecessarily wide, so I've instead reduced the right margin to 5px and set a minimum width on each line to ensure a mid width for things like the Math operator dropdown.

@rachel-fenichel FYI

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -252,6 +252,15 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
   var menu = new goog.ui.Menu();
   menu.setRightToLeft(this.sourceBlock_.RTL);
   for (var i = 0; i < options.length; i++) {
+    var separator = options[i] == 'SEPARATOR';
+    if (separator) {
+      // pxtblockly: render separator
+      var menuItem = new goog.ui.MenuSeparator();
+      menuItem.setRightToLeft(this.sourceBlock_.RTL);
+      menu.addChild(menuItem, true);
+      menuItem.getElement().style.borderColor = this.sourceBlock_.getColourTertiary();
+      continue;
+    }
     var content = options[i][0]; // Human-readable text or image.
     var value = options[i][1];   // Language-neutral value.
     if (typeof content == 'object') {
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -252,7 +252,9 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
   var menu = new goog.ui.Menu();
   menu.setRightToLeft(this.sourceBlock_.RTL);
   for (var i = 0; i < options.length; i++) {
-    var separator = options[i] == 'SEPARATOR';
+    var content = options[i][0]; // Human-readable text or image.
+    var value = options[i][1];   // Language-neutral value.
+    var separator = value === 'SEPARATOR';
     if (separator) {
       // pxtblockly: render separator
       var menuItem = new goog.ui.MenuSeparator();
@@ -261,8 +263,6 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
       menuItem.getElement().style.borderColor = this.sourceBlock_.getColourTertiary();
       continue;
     }
-    var content = options[i][0]; // Human-readable text or image.
-    var value = options[i][1];   // Language-neutral value.
     if (typeof content == 'object') {
       // An image, not text.
       var image = new Image(content['width'], content['height']);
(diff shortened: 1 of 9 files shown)


pelikhan:
The button is called “Make a variable” in the toolbox. Can we have the same terminology?

samelhusseini:
Yeah that would be the expected behaviour here. This PR is on pxt-blockly, and we override the Msg.NEW_VARIABLE in pxt to be "Make a variable" so ultimately that's what you'll see after we integrate.

pelikhan:
Perfect

core/field_variable.js
```diff
@@ -316,6 +316,8 @@ Blockly.FieldVariable.dropdownCreate = function() {
     // Set the UUID as the internal representation of the variable.
     options[i] = [variableModelList[i].name, variableModelList[i].getId()];
   }
+  options.push('SEPARATOR');
+  options.push([Blockly.Msg.NEW_VARIABLE, 'CREATE_VARIABLE']);
```
rachel-fenichel: Consider adding a constant (Blockly.CREATE_VARIABLE_ID?) instead of using a string directly here.

core/field_variable.js
```diff
@@ -348,6 +350,14 @@ Blockly.FieldVariable.prototype.onItemSelected = function(menu, menuItem) {
       // Delete variable.
       workspace.deleteVariableById(this.variable_.getId());
       return;
+    } else if (id == 'CREATE_VARIABLE') {
+      // Create a variable.
+      var that = this;
+      Blockly.Variables.createVariableButtonHandler(workspace, function(text) {
+        var variable = workspace.getVariable(text);
```
rachel-fenichel: What is the type of the variable?

Consider only having this option if the variable field can only hold a single type, and set the type of the new variable to the only type allowed by the field.

rachel-fenichel (commented):
I'll chat with the team about whether we want this, and I'll likely pull in the padding change regardless.

core/field_variable.js
samelhusseini: Right, yes that doesn't cover the dynamic variable scenario. I'll make it only show for variables of no type for now.

core/field_variable.js
samelhusseini: Yeah that makes sense. I was just being lazy haha. FYI https://github.com/google/blockly/blob/master/core/variables.js#L154 doesn't use a constant either.

samelhusseini:
@rachel-fenichel in regards to the padding change, I only changed it for blocklyDropdownDiv .goog-menuitem but I think the same can be applied for blocklyWidgetDiv, ie: also the context menu.
Although I haven't done that here yet.

core/field_dropdown.js
```diff
@@ -252,6 +252,15 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
   var menu = new goog.ui.Menu();
   menu.setRightToLeft(this.sourceBlock_.RTL);
   for (var i = 0; i < options.length; i++) {
+    var separator = options[i] == 'SEPARATOR';
```
samelhusseini: @rachel-fenichel I'm not a huge fan of this, do you have any recommendations here or does this seem fair to you?

core/field_variable.js
rachel-fenichel: I'll fix that :)

core/field_dropdown.js
rachel-fenichel: Interesting.  Have you tried using a separator in a non-variable dropdown?  I suspect it would break because of the checks in validateOptions.  Having it be a length-two array is probably better, even if we never use the first value.

samelhusseini: I have not, but yeah that makes sense. I'll fix that up.

core/field_variable.js
samelhusseini: Not sure if this is a problem with the dynamic_variables stuff but I would have thought this.defaultType_ would be set to non null in a dynamic variable. Tried Blockly mainline and same issue there. Is my assumption wrong? 
I would have to resort to reading the block type to figure out if I should show this option otherwise.

rachel-fenichel: Is it null or the empty string? [This](https://github.com/google/blockly/blob/master/core/field_variable.js#L255) code seems to indicate it would be the empty string.

Checking the block type is not the right move because fields can be on any type of block.  You can check the variableTypes array instead.

samelhusseini: this.defaultType_ is empty string and this.variableTypes is null, even in Blockly mainline typed variable playground.

rachel-fenichel (approved):

PR #180, opened 2018-12-17
Arg reporters

guillaumejenkins:
This PR adds argument reporter blocks that can be used instead of draggable variables for handler arguments.

Argument reporters are similar in behavior to draggable variables, but have some differences:
- They are not variables, so they will not appear in the workspace variables.
- They cannot be renamed.
- They have a type that sets their output shape and check.
- There are 4 types of argument reporter blocks: number, boolean, string and custom.
- Custom argument reporters have a mutation specifying the exact type of the argument.
- They cannot exist outside the event handler that defines them.
  - This means if they won't be connectable to any block that is not part of the handler's stack, unless the destination block is inside a different handler that defines an argument reporter of the same name and type.
  - It also means they will be deleted from their parent if you drag the parent from the handler's stack to a different block outside the handler, unless the destination is another handler that defines an argument reporter of the same name and type.
- No block can be connected on top of them on the handler block that defines them

These reporters will also be used for functions with parameters in a separate change.


blocks/functions.js
```diff
@@ -0,0 +1,230 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * @fileoverview Utility functions for handling functions (pxt-blockly's custom procedures).
+ * @author Microsoft MakeCode
```
samelhusseini: Your name and email here, format: 
```
 * @author samelh@microsoft.com (Sam El-Husseini)
```

core/functions.js
```diff
@@ -0,0 +1,91 @@
...
+ * Determines whether the specified type is custom or a built-in literal.
+ * @param {string} argumentType The argument type to check,
+ * @return {boolean} Whether the argument type is a custom type. A return value
+ *  of false means the argument is a built-in literal.
+ */
+Blockly.Functions.isCustomType = function (argumentType) {
+  return !(argumentType === 'boolean' ||
+    argumentType === 'string' ||
+    argumentType === 'number');
+}
```
samelhusseini: semi-colon.

core/functions.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {!Blockly.BlockSvg} block The block that should be used to make this
+ *     decision.
+ * @return {boolean} True if the block is a function argument reporter.
+ */
+Blockly.Functions.isFunctionArgumentReporter = function (block) {
+  return block.type === 'argument_reporter_boolean' ||
+    block.type === 'argument_reporter_number' ||
+    block.type === 'argument_reporter_string' ||
+    block.type === 'argument_reporter_custom';
+};
```
samelhusseini: new line

core/pxt_blockly_events.js
```diff
@@ -0,0 +1,90 @@
...
+ * @constructor
+ */
+Blockly.Events.EndBlockDrag = function(block) {
+  if (!block) {
+    return;  // Blank event to be populated by fromJson.
+  }
+  Blockly.Events.EndBlockDrag.superClass_.constructor.call(this, block);
+  this.recordUndo = false;
+  this.blockId = block.id;
+  this.allNestedIds = block.getDescendants().map(b => b.id);
```
samelhusseini: don't use arrow functions for IE compatibility.

core/pxt_blockly_utils.js
```diff
@@ -49,6 +49,38 @@ Blockly.pxtBlocklyUtils.measureText = function(fontSize, fontFamily,
...
+ * Finds and returns an argument reporter of the given name, argument type name, and reporter type on the
+ * given block, or null if none match.
+ * @param {!Blockly.Block} targetBlock The block to search.
+ * @param {!Blockly.Block} reporter The reporter to try to match.
+ * @return {boolean} Whether there is a matching reporter or not.
+ */
+Blockly.pxtBlocklyUtils.hasMatchingArgumentReporter = function(targetBlock, reporter) {
+  var argName = reporter.getFieldValue('VALUE');
+  var argTypeName = reporter.getTypeName();
+  var reporterType = reporter.type;
```
samelhusseini: unused variable

gh-pages/index.md
```diff
@@ -1,2 +1,2 @@
 * [Playground](playgrounds/tests/playground.html)
-* [Multi Playground](playgrounds/tests/multi_playground.html)
+* [Multi Playground](playgrounds/tests/multi_playground.html)
```
samelhusseini, line 2: Add your playground to this list.

gulpfile.js
```diff
@@ -76,6 +76,8 @@ gulp.task('publishall', ['python-build-all'], pxtPublishTask);
 
 gulp.task('publishts', [], pxtPublishTsTask);
 
+gulp.task('only-publishall', [], pxtPublishTask);
```
samelhusseini: More descriptive name or comment? What does this do?

samelhusseini:
Please add a PR description, makes it easier to look back on changes made.

blocks/functions.js
```diff
@@ -0,0 +1,230 @@
...
+// Argument editor and reporter helpers
+
+/**
+ * Returns the TypeScript type associated with this reporter or editor.
+ * @return {string} This argument's type, as would be emitted to TypeScript.
+ * @this Blockly.Block
+ */
+Blockly.PXTBlockly.FunctionUtils.getTypeName = function () {
+  return this.typeName_;
+}
```
samelhusseini: same: ; after function def

blocks/functions.js
```diff
@@ -0,0 +1,230 @@
...
+'use strict';
+
+/**
+ * Type to represent a function parameter
+ * @typedef {Object} FunctionParameter
+ * @property {string} id The blockly ID of the param
+ * @property {string} name the name of the param
+ * @property {string} type the type of the param (string, number, boolean or a custom type)
+ */
+
```
samelhusseini, line 45: nit: remove new line

gh-pages/index.md
guillaumejenkins: No new playground for reporter stuff, new playground will be added for functions

samelhusseini:
Install eslint and fix all the issues there.

gh-pages/index.md
samelhusseini: Oh yes.

blocks/functions.js
guillaumejenkins: The JSDoc is not related to the object being defined below, it just declares a type so the other JSDoc in the file can use that type

I fixed by moving the Utils declaration above the JSDoc

core/connection.js
```diff
@@ -315,11 +314,11 @@ Blockly.Connection.prototype.canConnectWithReason_ = function(target) {
   if (this.isSuperior()) {
     var blockA = this.sourceBlock_;
     var blockB = target.getSourceBlock();
-    var superiorConn = this;
+    // var superiorConn = this;
```
samelhusseini: Why are these commented out. If this is intended, could you add a comment describing what you're doing here?

samelhusseini (approved):

core/connection.js
guillaumejenkins: This variable was used in a check further down that we disabled in pxt-blockly. I did not want to delete the variable, since we did not delete the check that was using it (we only commented it out).

I'll add a comment

PR #185, opened 2019-01-16
Allow inline function renaming on function definitions

guillaumejenkins:
Function names on function definition blocks now use a text input instead of a label, allowing renaming of functions without going into the function editor.

Key changes:
- When rebuilding a function block, preserve all non-value inputs (only recreate arg inputs)
- Use the text input's validator to update a function's name and propagate it to all its callers


samelhusseini (approved):
We should consider using the scratch blocks style labels for these names to differentiate between function name and parameter names in the function edit dialog.

PR #191, opened 2019-01-25
Debugger - adding type definition for getting user variables.

gushuro:
We need this for the debugger, for filtering only user-created variables.

### DIFF
--- a/typings/blockly.d.ts
+++ b/typings/blockly.d.ts
@@ -1061,6 +1061,7 @@ declare namespace Blockly {
         let flyoutCategory: (wp: Workspace) => HTMLElement[];
         let flyoutCategoryBlocks: (wp: Workspace) => HTMLElement[];
         function createVariable(wp: Workspace, opt_callback?: ((e: any) => void)): void;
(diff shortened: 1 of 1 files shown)


samelhusseini (approved):

typings/blockly.d.ts
```diff
@@ -1061,6 +1061,7 @@ declare namespace Blockly {
         let flyoutCategory: (wp: Workspace) => HTMLElement[];
         let flyoutCategoryBlocks: (wp: Workspace) => HTMLElement[];
         function createVariable(wp: Workspace, opt_callback?: ((e: any) => void)): void;
+        function allUsedVarModels(ws: Blockly.Workspace): Array<Blockly.VariableModel>;
```
riknoll: For arrays, we typically use the syntax:
```typescript
Blockly.VariableModel[]
```
instead of the generic angle-brackets

riknoll (changes requested):

typings/blockly.d.ts
gushuro: Cool, changed.

riknoll (approved):

PR #193, opened 2019-01-28
Fix typo

changeworld:
ture -> true


samelhusseini:
Thanks for the PR, however we actually don't use these localization files in PXT, and actually translate all of our blocks on crowdin. see: makecode.com/translate. 

The Google/Blockly team don't translate their files directly either (ie: through PRs) and instead use a service called translatewiki. see: https://developers.google.com/blockly/guides/modify/web/translating if you're interested in contributing to that.

Thanks again!

PR #194, opened 2019-01-28
Fade blocks with outputs while dragging

rachel-fenichel:
Per discussion with @samelhusseini.  You'll want to play with the opacity, I think.

This sets the opacity on the drag surface instead of on the block, so that it applies equally to all child blocks.

### DIFF
--- a/core/insertion_marker_manager.js
+++ b/core/insertion_marker_manager.js
@@ -154,6 +154,8 @@ Blockly.InsertionMarkerManager = function(block, handleXY) {
   if (block.outputConnection) {
     var coord = new goog.math.Coordinate(block.outputConnection.x_, block.outputConnection.y_);
     this.handleDXY = goog.math.Coordinate.difference(handleXY, coord);
+    // Fade the dragging block to make targets easier to see.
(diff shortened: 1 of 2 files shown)


pelikhan (approved):

core/insertion_marker_manager.js
```diff
@@ -154,6 +154,8 @@ Blockly.InsertionMarkerManager = function(block, handleXY) {
   if (block.outputConnection) {
     var coord = new goog.math.Coordinate(block.outputConnection.x_, block.outputConnection.y_);
     this.handleDXY = goog.math.Coordinate.difference(handleXY, coord);
+    // Fade the dragging block to make targets easier to see.
+    this.workspace_.getBlockDragSurface().setOpacity(0.7);
```
pelikhan, line 158: should this be a constant we can change?

riknoll (approved):

core/insertion_marker_manager.js
riknoll: I think it's fine to hardcode it, we won't be changing it between editors I don't think.

PR #196, opened 2019-02-01
Support semantic-like icons for function args

guillaumejenkins:
Changes the arg icon mechanism for the function editor from base-64 encoded SVGs inside an `` element, to `<i>` elements using semantic-style classNames

### DIFF
--- a/core/field_argumenteditor.js
+++ b/core/field_argumenteditor.js
@@ -85,11 +85,11 @@ Blockly.FieldArgumentEditor.prototype.showEditor_ = function() {
   div.appendChild(removeButton);
 
   if (this.sourceBlock_ && this.sourceBlock_.typeName_) {
-    var typeIconSvgData =
+    var iconClass =
         Blockly.PXTBlockly.FunctionUtils.getArgumentIcon(this.sourceBlock_.typeName_);
-    if (typeIconSvgData) {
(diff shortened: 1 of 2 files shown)


core/field_argumenteditor.js
```diff
@@ -85,11 +85,11 @@ Blockly.FieldArgumentEditor.prototype.showEditor_ = function() {
...
   if (this.sourceBlock_ && this.sourceBlock_.typeName_) {
-    var typeIconSvgData =
+    var iconClass =
         Blockly.PXTBlockly.FunctionUtils.getArgumentIcon(this.sourceBlock_.typeName_);
-    if (typeIconSvgData) {
-      var typeIcon = goog.dom.createDom(goog.dom.TagName.IMG, 'argumentEditorTypeIcon');
-      typeIcon.setAttribute('src', typeIconSvgData);
+    if (iconClass) {
+      var className = iconClass + ' icon argumentEditorTypeIcon';
+      var typeIcon = goog.dom.createDom(goog.dom.TagName.I, className);
```
riknoll: I believe the Blockly folks are trying to remove all of the `goog.dom` dependencies, so it's probably best to use `document.createElement()` in new code

riknoll (approved):

core/field_argumenteditor.js
guillaumejenkins: ah, thanks for the info

PR #198, opened 2019-02-08
Dropdown field refactoring

rachel-fenichel:
This change breaks the `showEditor` function into a few helper functions that match the ones in Blockly's dropdown field.  As part of rendering I'm teasing out the common code and understanding the code that is different between the two renderers.

Tested in the playground.


rachel-fenichel:
Friendly ping.

riknoll (approved):

PR #202, opened 2019-02-20
Make all shadow blocks duplicate on drag

riknoll:
This PR makes all shadow blocks duplicate when dragged. It's the same behavior that we use for function arguments, but applied to all shadow blocks and not just variables. The motivation for this change is to make refactoring and duplicating code easier in blocks; we've gotten a lot of feedback in MakeCode Arcade that images are especially hard to copy between blocks because it's non-trivial to redraw them in the field editor.
duplicate_on_drag

I'm making this a draft pull request because I think this probably requires some discussion before merging.

@rachel-fenichel do you see any issues this creates that I might be overlooking? I'm curious if the Blockly team has considered this approach before.

### DIFF
--- a/core/pxt_blockly_utils.js
+++ b/core/pxt_blockly_utils.js
@@ -49,12 +49,7 @@ Blockly.pxtBlocklyUtils.measureText = function(fontSize, fontFamily,
  * @package
  */
 Blockly.pxtBlocklyUtils.isShadowArgumentReporter = function(block) {
-  return block.isShadow() &&
-    (block.type === 'variables_get_reporter' ||
-      block.type === 'argument_reporter_boolean' ||
-      block.type === 'argument_reporter_number' ||
-      block.type === 'argument_reporter_string' ||
-      block.type === 'argument_reporter_custom');
+  return block.isShadow()
 };
 
 /**


samelhusseini:
One issue you might run into here is if a user intends to drag the parent block and just happens to have clicked on the shadow, since shadow blocks appear to be part of the parent block. You could potentially get the same effect (ease of refactoring) but instead have the user hold down the ctrl / meta key.

abchatra (approved):
Looks good to me.

riknoll:
The meta-key doesn't really address touch though (where copy/paste is especially difficult)

samelhusseini:
For touch maybe you could add a button that enables “copy/duplicate mode”, then the next thing you touch can be duplicated including shadow blocks. Just a thought

rachel-fenichel:
Interesting.  I haven't played with this idea before.  I agree with what Sam has said.

abchatra:
I like making it a default option. Drag it and it duplicates the shadow block.

darzu:
having a copy/duplicated mode sounds confusing to me. Modes are really hard to keep straight IMO, it's easy to forget which mode you're in. Also it doesn't seem very discoverable. So far all the modes I know of in MakeCode come with very noticeable visual changes like blocks<->typescript, full screen simulator, serial console, and the debugger.

darzu (approved):

Issue #206, opened 2019-03-08
Dropdown crash


pelikhan:
Seeing some crashes related to https://github.com/Microsoft/pxt-blockly/commit/9f61a3f6d1964dfaff9dee7c8b9330f6a2d51a29 in https://maker.makecode.com . This breaks every dropdown field.

```
pxtblockly.js:formatted:30220 Uncaught TypeError: this.createMenu_ is not a function
    at FieldNumberDropdown.Blockly.FieldDropdown.showEditor_ [as showDropdown_] (pxtblockly.js:formatted:30220)
    at FieldNumberDropdown.<anonymous> (pxtblockly.js:formatted:30651)
    at FieldNumberDropdown.Blockly.FieldTextInput.showEditor_ (pxtblockly.js:formatted:29487)
    at FieldNumberDropdown.Blockly.FieldTextDropdown.showEditor_ (pxtblockly.js:formatted:30650)
    at Blockly.TouchGesture.Blockly.Gesture.doFieldClick_ (pxtblockly.js:formatted:22842)
    at Blockly.TouchGesture.Blockly.Gesture.handleUp (pxtblockly.js:formatted:22788)
    at Blockly.TouchGesture.handleUp (pxtblockly.js:formatted:22987)
    at HTMLDocument.h (pxtblockly.js:formatted:33589)
```

pelikhan:
@rachel-fenichel @riknoll  FYI

rachel-fenichel:
GTK (whoops)

PR #207, opened 2019-03-08
Reverting dropdown menu refactoring

pelikhan:
Fix for https://github.com/Microsoft/pxt-blockly/issues/206 . I'm reverting the refactoring so that we get more time in the future to re-apply /test it.


pelikhan:
@rachel-fenichel FYI

PR #208, opened 2019-03-11
Support for multiple flyouts; additional typing; scrollTo perf improvement

darzu:
RELATED: https://github.com/Microsoft/pxt/pull/5323

- Adds helper function for creating & initializing a flyout which allows PXT to maintain multiple flyouts per toolbox (roughly one per category). Maintaining multiple flyouts drastically improves performance (~3X) when switching between categories.
- Adds additional private typings for maintaining multiple flyouts in pxt-core (we violate encapsulation)
- Short-circuit scrollbar's scrollTo if the destination is the same as the current position, this improves performance


pelikhan:
Make sure to ingest https://github.com/Microsoft/pxt-blockly/pull/207

pelikhan (approved):

typings/blockly.d.ts
```diff
@@ -13422,6 +13436,14 @@ declare module Blockly.Functions {
...
      */
     function isFunctionArgumentReporter(block: Blockly.BlockSvg): boolean;
+
+    /**
+     * Create a flyout, creates the DOM elements for the flyout, and initializes the flyout.
+     * @param {!Blockly.Workspace} workspace The target and parent workspace for this flyout. The workspace's options will
+     *     be used to create the flyout's inner workspace.
+     * @return { flyout: !Blockly.Flyout, flyoutSvg: SVGSVGElement } True if the block is a function argument reporter.
+     */
+    function createFlyout(workspace: Blockly.Workspace): void;
```
pelikhan: missing "return" type? did you use "gulp generate-dts" to regen blockly.d.ts?

darzu: Looks like generate-dts cannot handle record types. I'll find a workaround.

pelikhan (approved):

darzu:
@riknoll @pelikhan , who was the Blockly team member I should @ ?

pelikhan:
@rachel-fenichel

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+// Place your settings in this file to overwrite default and user settings.
```
pelikhan, line 1: not sure we want this file

core/pxt_blockly_functions.js
```diff
@@ -640,3 +640,34 @@ Blockly.Functions.isFunctionArgumentReporter = function(block) {
...
+
+/**
+ * Create a flyout, creates the DOM elements for the flyout, and initializes the flyout.
+ * @param {!Blockly.Workspace} workspace The target and parent workspace for this flyout. The workspace's options will
+ *     be used to create the flyout's inner workspace.
+ * @param {!Element} siblingNode The flyout is added after this reference node. 
+ * @return {!Blockly.Flyout} The newly created flyout.
+ */
+Blockly.Functions.createFlyout = function (workspace, siblingNode) {
+  let flyoutWorkspaceOptions = {
```
pelikhan, line 652: use var

core/toolbox.js
```diff
@@ -175,29 +175,12 @@ Blockly.Toolbox.prototype.init = function() {
...
-    disabledPatternId: workspace.options.disabledPatternId,
-    parentWorkspace: workspace,
-    RTL: workspace.RTL,
-    oneBasedIndex: workspace.options.oneBasedIndex,
-    horizontalLayout: workspace.horizontalLayout,
-    toolboxPosition: workspace.options.toolboxPosition,
-    // pxt-blockly: pass the newFunctions option
-    newFunctions: workspace.options.newFunctions
-  };
+  const flyout = Blockly.Functions.createFlyout(workspace, this.workspace_.getParentSvg());
```
pelikhan, line 178: use var

pelikhan (approved):
Test these changes in IE.

darzu:
> Test these changes in IE.

Crap, just saw this. My github isn't auto-refreshing PRs

Will test now.

pelikhan:
IE does not support let/const. it’s easy to leak one in blockly.

darzu:
We should have a compile check for that

darzu:
Looks like "let" partially works on IE. It's basically equivalent to var: https://caniuse.com/#search=let

It's a huge pain that github let me hit "merge" while the comments weren't refreshed. That's something VSTS/Azure DevOps got right: you could set a policy that all comments needed to be "resolved" before check in. Made it much easier to be sure you weren't missing something. </rant>

darzu:
Tested in IE. Looks like everything works, and the perf improvement applies there as well.

.vscode/settings.json
darzu: it's a pain not having this if you have formatOnSave enabled by default since blockly seems to use some formatting settings vscode doesn't understand.
Should be safe to check this in, no?

pelikhan:
And cleanup your branches ;)

PR #212, opened 2019-03-13
Remove destructive side-effects from Flyout.hide()

darzu:
When maintaining multiple flyouts for a single toolbox, it's important that we don't destruct (parts of) the flyout when calling hide().

Previously, hide() was deleting all event listeners from the flyout, but this breaks the case where we just want to toggle the visibility of the flyout instead of clearing & recreating the flyout each time we open and close it. So instead I've moved the event listener cleanup to the callers of hide() that where depending on that cleanup behavior.

Risks:
- Perhaps some of these event listeners will trigger in undesirable ways even when the flyout is closed (b/c previously these events where unregistered each time we hid the flyout)

Mitigations:
- [x] instrument all event listeners on a flyout and determine if any will fire when the flyout is closed with manual testing

Todo:
- [ ] Add `// pxt-blockly: <reason>` comments explaining changes made for easier merges from upstream

Part of a fix for:
https://github.com/Microsoft/pxt/pull/5337


core/flyout_base.js
```diff
@@ -397,14 +398,20 @@ Blockly.Flyout.prototype.positionAt_ = function(width, height, x, y) {
...
 /**
- * Hide and empty the flyout.
+ * Hide the flyout.
  */
 Blockly.Flyout.prototype.hide = function() {
   if (!this.isVisible()) {
     return;
   }
   this.setVisible(false);
-  // Delete all the event listeners.
```
pelikhan, line 407: Could you add comments like “event listeners are not deleted!”

pelikhan (approved):

pelikhan (approved):

pelikhan (approved):

PR #214, opened 2019-03-16
Adding debugMode Option

gushuro:
Added debugMode to workspace options. Updated movability and editability
of blocks accordingly.

Also fixed two small bugs with breakpoints.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -649,7 +649,7 @@ Blockly.Block.prototype.setDeletable = function(deletable) {
  * @return {boolean} True if movable.
  */
 Blockly.Block.prototype.isMovable = function() {
-  return this.isMovablePersisted() && !this.workspace.options.debugMode;
+  return this.isMovablePersisted() && !(this.workspace && this.workspace.options.debugMode);
 };
 
 /**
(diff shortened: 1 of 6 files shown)


core/block.js
```diff
@@ -650,7 +650,17 @@ Blockly.Block.prototype.setDeletable = function(deletable) {
  */
 Blockly.Block.prototype.isMovable = function() {
   return this.movable_ && !this.isShadow_ &&
-      !(this.workspace && this.workspace.options.readOnly);
+    !(this.workspace &&
```
riknoll: Should probably just call `isMovablePersisted()` instead of duplicating the logic. Same with `isEditable()` below

core/workspace.js
```diff
@@ -662,6 +662,14 @@ Blockly.Workspace.getById = function(id) {
   return Blockly.Workspace.WorkspaceDB_[id] || null;
 };
 
+/**
+ * Sets the debugMode option in the workspace.
+ * @param {boolean} debugMode value to set to this option.
+ */
+Blockly.Workspace.prototype.setDebugmodeOption = function(debugMode) {
```
riknoll: Capitalize the M

riknoll (changes requested):

core/workspace.js
gushuro: Sure

PR #218, opened 2019-04-01
Debugger: don't show variables on disabled blocks

gushuro:
Making AllUsedVarModels return only variables defined by the user in non-disabled blocks.


riknoll (approved):
It would be better if we could do this in PXT but it's such a small change that I think it's fine.

PR #224, opened 2019-04-03
Reconnect child blocks after undo/redo, fix input rendering for if-else

shakao:
Fix undo/redo issues (Microsoft/pxt-microbit/issues/1486) with if-else block 

undo-ifelse


abchatra (approved):
Looks good to me. @riknoll probably needs to take a look as well

riknoll (approved):

PR #226, opened 2019-04-09
Switch string fields to use text quotes instead of SVGs

ChaseMor:
This changes the quotes used in the string field to display as text characters instead of svgs. 

Switched for better compatibility across browsers.

Fixes Microsoft/pxt-microbit#1710

## Before
image

## After
image


core/css.js
```diff
@@ -1286,5 +1286,10 @@ Blockly.Css.CONTENT = [
     'font-weight: bold;',
   '}',
 
+  // Quote for string field
+  '.text-quote {',
```
riknoll: Could this be a bit more descriptive? Maybe `field-text-quote`?

riknoll (approved):

riknoll (approved):

PR #233, opened 2019-04-16
Added adjustments to bounding box if block has hat

### LINKED ISSUE #1950: Beta: hat missing when downloading snapshot of function
This is the snapshot downloaded from makecode.microbit.org/beta of a function with parameters

microbit-screenshot(4)

This is the equivalent screenshot
Screenshot 2019-04-03 at 08 49 40

Expected:
hat should be included as part of the function when downloading the snapshot

ChaseMor:
Checks if a block has a hat, if they do adjust the bounding box by bumping the top-left corner to the top of the hat.

Fixes Microsoft/pxt-microbit#1950

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -626,7 +626,9 @@ Blockly.BlockSvg.prototype.getBoundingRectangle = function() {
     bottomRight = new goog.math.Coordinate(blockXY.x + blockBounds.width,
         blockXY.y + blockBounds.height);
   }
-
+  if (this.startHat_) {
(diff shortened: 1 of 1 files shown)


pelikhan (approved):
Nice. Any visible side effects?

ChaseMor:
The only thing that looks different is that when selecting ``format code`` the top of non-hat blocks line up with the top of the hat for other blocks

image


where as previously, it would look like

image

pelikhan:
Good enough

PR #234, opened 2019-04-17
Use variable name instead of name+id as identifier

shakao:
Use variable name instead of name+id as identifier, to avoid conflicts with cached variables (Microsoft/pxt-microbit/issues/1968)

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -783,7 +783,8 @@ Blockly.Xml.domToFieldVariable_ = function(workspace, xml, text, field) {
     type = '';
   }
 
-  var variable = Blockly.Variables.getOrCreateVariablePackage(workspace, xml.id,
+  // pxt-blockly: variable ID and variable name are both unique, only use name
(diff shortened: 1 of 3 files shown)


core/xml.js
```diff
@@ -783,7 +783,8 @@ Blockly.Xml.domToFieldVariable_ = function(workspace, xml, text, field) {
     type = '';
   }
 
-  var variable = Blockly.Variables.getOrCreateVariablePackage(workspace, xml.id,
+  // pxt-blockly: variable ID and variable name are both unique, only use name
+  var variable = Blockly.Variables.getOrCreateVariablePackage(workspace, null,
```
pelikhan, line 787: can the variable name be empty?

riknoll (approved):

core/xml.js
riknoll: I just tried it on micro:bit and it doesn't seem like it

PR #235, opened 2019-04-18
Add collapse/expand to right-click menu if enabled in target

shakao:
Add collapse/expand to right-click menu if enabled in target (Microsoft/pxt-arcade/issues/302)

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -792,6 +792,28 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
     if (this.isEditable() && this.workspace.options.comments) {
       menuOptions.push(Blockly.ContextMenu.blockCommentOption(block));
     }
+
+    // pxt-blockly: only allow collapse on top-level event handler blocks
+    if (this.workspace.options.collapse
+        && block.previousConnection == null && block.outputConnection == null) {
+      // Option to collapse/expand block.
+      if (this.collapsed_) {
+        var expandOption = {enabled: true};
+        expandOption.text = Blockly.Msg['EXPAND_BLOCK'];
+        expandOption.callback = function() {
+          block.setCollapsed(false);
+        };
+        menuOptions.push(expandOption);
+      } else {
+        var collapseOption = {enabled: true};
+        collapseOption.text = Blockly.Msg['COLLAPSE_BLOCK'];
+        collapseOption.callback = function() {
+          block.setCollapsed(true);
+        };
+        menuOptions.push(collapseOption);
+      }
+    }
+
     menuOptions.push(Blockly.ContextMenu.blockDeleteOption(block));
   } else if (this.parentBlock_ && (this.isShadow_
       && !Blockly.pxtBlocklyUtils.isShadowArgumentReporter(block))) {


abchatra:
Can you attach a gif? @Jaqster

core/block_svg.js
```diff
@@ -792,6 +792,28 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
     if (this.isEditable() && this.workspace.options.comments) {
       menuOptions.push(Blockly.ContextMenu.blockCommentOption(block));
     }
+
+    // pxt-blockly: only allow collapse on top-level event handler blocks
+    if (this.workspace.options.collapse
+        && block.previousConnection == null && block.outputConnection == null) {
```
pelikhan, line 798: can previousConnection be undefined?

abchatra: undefined == null is true right?

pelikhan: i find !block.previousConnection less confusing.

core/block_svg.js
```diff
@@ -792,6 +792,28 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
...
+      // Option to collapse/expand block.
+      if (this.collapsed_) {
+        var expandOption = {enabled: true};
+        expandOption.text = Blockly.Msg['EXPAND_BLOCK'];
```
pelikhan, line 802: we need to localize those strings.

core/block_svg.js
```diff
@@ -792,6 +792,28 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
...
+        expandOption.callback = function() {
+          block.setCollapsed(false);
+        };
+        menuOptions.push(expandOption);
+      } else {
+        var collapseOption = {enabled: true};
+        collapseOption.text = Blockly.Msg['COLLAPSE_BLOCK'];
```
pelikhan, line 809: localize this one too

pelikhan (approved):

core/block_svg.js
shakao: these strings were already in the Blockly.Msg dictionary, and it looks like they've been localized for most languages

shakao:
collapse

Collapse/expand feature @Jaqster

core/block_svg.js
pelikhan: We do not use the blockly localized strings — there is a place (maybe in pxt) where we redefined those using the LF macro.

shakao: just checked and they've already been localized in pxt as well--it seems like we've localized all the context menu items (like "Inline Inputs" and "External Inputs" are also localized) so that's nice!

riknoll (approved):

abchatra (approved):

darzu:
Neat feature but I worry it'll be confusing. Advanced users might use this a lot but when other users try to remix this, it might not be obvious what's going on. It's hard to tell a collapsed block from just a regular block, and it's hard to tell how to uncollapse the block. Lots of users struggle with right-click.

abchatra:
This is requested by lot of hero teachers. If someone has right clicked to collapse, thinking was they will know how to expand. This should be part of undo stack as well. 

Although I see your point. We should take this to UI design meeting. May be + icon to add at the end to expand will suffice.

pelikhan:
expand/collapse should be on by default on functions in makecode -- to encourage abstraction.

pelikhan:
Can we get this in?

abchatra:
This is stale PR. We need to implement the new design.

PR #242, opened 2019-06-08
fix crash when same variable is cached in multiple flyouts

jwunderl:
Fixes https://github.com/microsoft/pxt-arcade/issues/980

My understanding is we don't need this crash condition, cause we already only allow a single variable per variable name (versus default blockly which can have multi variables of different types with the same name). I can look into fixing this by invalidating the cached flyouts or something if that approach is preferred, but I don't think this is an an error that really applies to us. Holding off on changing any tests till that's confirmed though.

One other thing: I needed to ``npm link`` blockly from pxt for my changes to apply, not sure if that's intentional (I didn't see it mentioned anywhere in the docs, can add it).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ node_js:
 sudo: required
 addons:
   chrome: stable
-  apt:
-    packages:
-    - libgconf-2-4
 cache:
   directories:
   - node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+os: linux
+dist: trusty
 node_js:
 - "8"
 sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ node_js:
 sudo: required
 addons:
   chrome: stable
+  apt:
+    packages:
+    - libgconf-2-4
 cache:
   directories:
   - node_modules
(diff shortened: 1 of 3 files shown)


jwunderl:
Looks like build is failing due to recent travis update that went from ubuntu14->16.04. Updated `.travis.yml` to try bundling `libgconf` that is no longer included by default, might need some more work? re: https://github.com/cypress-io/cypress/issues/4069#issuecomment-488315675

jwunderl:
Fixed build by pegging it to use 14 ('trusty') instead of the new one (noticed that it was possible from https://github.com/google/blockly/pull/2403)

.travis.yml
```diff
@@ -1,9 +1,14 @@
...
+os: linux
+dist: trusty
 node_js:
 - "8"
 sudo: required
 addons:
   chrome: stable
+  apt:
+    packages:
+    - libgconf-2-4
```
shakao: is this still necessary with the "trusty" environment?

shakao (approved):

core/variable_map.js
```diff
@@ -173,12 +173,7 @@ Blockly.VariableMap.prototype.createVariable = function(name,
     opt_type, opt_id) {
   var variable = this.getVariable(name, opt_type);
   if (variable) {
-    if (opt_id && variable.getId() != opt_id) {
-      throw Error('Variable "' + name + '" is already in use and its id is "' +
-                  variable.getId() + '" which conflicts with the passed in ' +
-                  'id, "' + opt_id + '".');
-    }
-    // The variable already exists and has the same ID.
+    // pxt-blockly: no error if the wrong ID is passed, as we do not allow duplicate variable names
```
pelikhan, line 176: does blockly have some kind of options where we could disable this check, instead of erasing it. will that be a problem when trying to keep in sync with blockly maintream?

PR #244, opened 2019-06-20
New slider

ChaseMor:
Get's rid of the Google Closure Slider and replaces it with a new Slider object. It looks the same. Really the only functional difference is that if a value for the slider is greater than the max or less than the min, the thumb of the slider will show clamped on that side.

Fixes https://github.com/microsoft/pxt-arcade/issues/460


core/slider.js
```diff
@@ -0,0 +1,147 @@
+/**
+ * @license
```
riknoll, line 2: Wrong license

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview The slider class used for field sliders
+ * @author Chase Mortensen
```
riknoll, line 13: we don't usually put authors in files

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+            if (this.slider.onClickEvent) {
+                this.slider.onClickEvent();
+            }
+            if (this.moveToPoint) {
+                this.updatePosition(ev.clientX);
+            }
+        }
+    })
+
+    Blockly.bindEvent_(window, "mouseup", this, function (ev) {
```
riknoll: If you bind an event on the window, it needs to get cleaned up somewhere. Typically we bind these events in the mousedown and mouseup handlers, so the flow looks something like this:

1. mousedown fires
2. In the event handler, register the mouseup and mousemove events on the window
3. mouseup fires
4. In the event handler, remove the mouseup and mousemove handlers from the window

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+
+    Blockly.bindEvent_(window, "mousemove", this, function(ev) {
+        if (this && this.slider && this.slider.mouseDown_) {
+            this.updatePosition(ev.clientX);
+        }
+    })
+}
+
+Blockly.Slider.prototype.updatePosition = function(clientX) {
+    const rect = (this.slider.element_).getBoundingClientRect();
```
riknoll: Don't use const and let in JavaScript files. Some browsers don't support them (TypeScript converts them all to vars)

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+    })
+
+    Blockly.bindEvent_(window, "mouseup", this, function (ev) {
+        if (this && this.slider && this.slider.mouseDown_) {
+            this.slider.mouseDown_ = false;
+        }
+    })
+
+    Blockly.bindEvent_(window, "mousemove", this, function(ev) {
+        if (this && this.slider && this.slider.mouseDown_) {
```
riknoll: I don't think you need to check `this` if you are using `Blockly.bindEvent_`

core/slider.js
riknoll: Also, parentheses

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+Blockly.Slider.prototype.setStep = function (step) {
+    this.step_ = step;
+};
+
+Blockly.Slider.prototype.setMoveToPointEnabled = function (enabled) {
+    this.moveToPoint = enabled;
+}
+
+Blockly.Slider.prototype.setValue = function (value) {
+    this.value_ = value;
```
riknoll, line 144: Shouldn't this be assigned after the min/max is applied?

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+Blockly.Slider.prototype.render = function (parentElement, opt_beforeNode) {
+    if (!this.element_) {
+        this.createDom();
+    }
+    parentElement.insertBefore(this.element_, opt_beforeNode || null);
+};
+
+Blockly.Slider.prototype.createDom = function () {
+    this.element_ = document.createElement('div');
+    this.element_.classList.add('slider');
```
riknoll: Can't use classList, IE doesn't support it. I think Blockly has a helper function for this

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+ * @author Chase Mortensen
+ */
+'use strict';
+
+goog.provide('Blockly.Slider');
+    
+Blockly.Slider = function() {
+
+    // Set up default values
+    this.step_ = 1;
```
riknoll: Blockly uses the prototype for setting default values. See https://github.com/microsoft/pxt-blockly/blob/develop/core/comment.js#L52

core/slider.js
```diff
@@ -0,0 +1,147 @@
...
+
+    this.slider = this;
```
riknoll: What is this for? If you can do `x.slider`, you must have a reference to `x`

riknoll (changes requested):
Please add JSDoc for all of the methods and properties. I know we don't in pxt proper, but we usually try to match the style in blockly

core/slider.js
ChaseMor: This is run when the user types in a value to the html field. We don't necessarily want to constrain that value, but we want to constrain the slider's position.

PR #245, opened 2019-08-27
Pull google/blockly changes through July 2019 release

shakao:
Major updates include:
- Changes to field editor APIs
- Rename of utility functions
- Theme APIs
- Beginning of work to remove closure

Additional details here: https://github.com/google/blockly/releases/

Most of the major changes from pxt-blockly are here: https://github.com/microsoft/pxt-blockly/pull/245/commits/562114736c2bd584725d3bd00ece2ea9008a1168


core/field_colour.js
```diff
@@ -127,6 +127,13 @@ Blockly.FieldColour.prototype.titles_ = null;
  */
 Blockly.FieldColour.prototype.columns_ = 0;
 
+/**
+ * The color picker.
+ * @type {!Element}
+ * @private
+ */
+Blockly.FieldColour.prototype.colorPicker_ = null;
```
riknoll, line 135: `colourPicker`?

core/field_textinput.js
```diff
@@ -374,12 +382,21 @@ Blockly.FieldTextInput.prototype.showInlineEditor_ = function(
  * @param {Function=} arrowCallback Callback for when drop-down arrow clicked.
  * @private
  */
-Blockly.FieldTextInput.prototype.widgetCreate_ = function(withArrow, arrowCallback) {
+Blockly.FieldTextInput.prototype.widgetCreate_ = function(
+    readOnly, withArrow, arrowCallback) {
   var div = Blockly.WidgetDiv.DIV;
-
+  // Apply text-input-specific fixed CSS
+  div.className += ' fieldTextInput';
```
riknoll, line 389: Can't use className in IE

core/variables.js
```diff
@@ -449,6 +449,27 @@ Blockly.Variables.nameUsedWithAnyType_ = function(name, workspace) {
...
+ */
+Blockly.Variables.generateVariableFieldXmlString = function(variableModel) {
+  // The variable name may be user input, so it may contain characters that need
+  // to be escaped to create valid XML.
+  var typeString = variableModel.type;
+  if (typeString == '') {
+    typeString = '\'\'';
+  }
+  // TODO shakao remove goog.string
+  var text = '<field name="VAR" id="' + variableModel.getId() +
```
riknoll, line 467: We have an htmlEscape function in PXT that you could steal

core/field_textinput.js
samelhusseini: I think you must be thinking of [classList](https://caniuse.com/#feat=classlist). 
className should be supported in IE, otherwise I think we're in trouble as we use it all over the place.

riknoll: Yep, my mistake!

riknoll (approved):
LGTM! I only looked at the post-merge commits

jwunderl (approved):

gulpfile.js
```diff
@@ -1,66 +1,185 @@
...
-		console.log('python exited with code ' + code);
-		cb(code);
-	});
-});
+////////////////////////////////////////////////////////////
+//                        PXT                             //
+////////////////////////////////////////////////////////////
 
-gulp.task("generate-dts", function(cb) {
-	console.info('Generate blockly.d.ts')
```
samelhusseini, line 55: I've moved the dts generation over to gulp (see ``typings`` task above). Since you're getting rid of the old task, you might also want to get rid of old generation script: https://github.com/microsoft/pxt-blockly/blob/develop/typings/generate-dts.sh.

package/README.md
```diff
@@ -0,0 +1,77 @@
+# Blockly
```
samelhusseini: It looks like you've left out the npm publish tasks out of the ``gulpfile.js``. In that case, you have no need for any of the files in the ``package`` directory. Just an FYI incase you were wondering what they were for.

core/field_colour.js
shakao: from before: https://github.com/microsoft/pxt-blockly/blob/develop/core/field_colour.js#L85 (will leave a note to rename in a future change)

package/README.md
shakao: removed them for now, thanks! i also took out our own npm publish stuff, i wanted to see if i could sync it with yours at a later point in time but will probably save that for a separate PR

core/workspace_svg.js
```diff
@@ -1269,43 +1422,38 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
...
+      boundary.right = blockBoundary.right;
     }
   }
-  return {
-    x: boundary.topLeft.x,
-    y: boundary.topLeft.y,
-    width: boundary.bottomRight.x - boundary.topLeft.x,
-    height: boundary.bottomRight.y - boundary.topLeft.y
-  };
+  return boundary;
```
samelhusseini, line 1456: Just a heads up. This is incompatible with what pxt is expecting here: 
https://github.com/microsoft/pxt/blob/master/pxtblocks/blocklylayout.ts#L216

You'll have to update the logic there to use: top, left, bottom, right instead.

shakao: good catch, thanks!

PR #252, opened 2019-10-08
handle 0 case in field-slider

pelikhan:
You probably want to test locally your changes against all editors: adafruit, maker, arcade, microbit. Each of them has slightly different blockly coverage.

### DIFF
--- a/core/field_slider.js
+++ b/core/field_slider.js
@@ -44,8 +44,8 @@ Blockly.FieldSlider = function(opt_value, opt_min, opt_max, opt_precision,
     opt_step, opt_labelText, opt_validator) {
   Blockly.FieldSlider.superClass_.constructor.call(this, opt_value, opt_min,
       opt_max, opt_precision, opt_validator);
-  this.min_ = parseFloat(opt_min) || -Infinity;
-  this.max_ = parseFloat(opt_max) || Infinity;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
Hm, setOptions on line 104 below looks suspicious too then? Including step / precision. Cant wait for nullish coalescing in however many years support for that takes 

Probably preferable to split the lines per statement

core/field_slider.js
```diff
@@ -44,8 +44,8 @@ Blockly.FieldSlider = function(opt_value, opt_min, opt_max, opt_precision,
     opt_step, opt_labelText, opt_validator) {
   Blockly.FieldSlider.superClass_.constructor.call(this, opt_value, opt_min,
       opt_max, opt_precision, opt_validator);
-  this.min_ = parseFloat(opt_min) || -Infinity;
-  this.max_ = parseFloat(opt_max) || Infinity;
+  this.min_ = parseFloat(opt_min); if (isNaN(this.min_)) this.min_ = -Infinity;
+  this.max_ = parseFloat(opt_max); if (isNaN(this.max)) this.max_ = Infinity;
```
samelhusseini, line 48: Could you perhaps use ``this.setConstraints_(...)`` instead of setting each of these directly.

PR #259, opened 2019-10-18
Use document.createElement in some cases, update callbackKey name

shakao:


jwunderl (approved):

core/pxt_blockly_functions.js
```diff
@@ -110,14 +110,14 @@ Blockly.Functions.flyoutCategory = function(workspace) {
...
-  var button = Blockly.utils.xml.createElement('button');
+  var button = document.createElement('button');
   var msg = Blockly.Msg.FUNCTION_CREATE_NEW;
   var callbackKey = 'CREATE_FUNCTION';
   var callback = function() {
     Blockly.Functions.createFunctionCallback_(workspace);
   };
   button.setAttribute('text', msg);
-  button.setAttribute('callbackkey', callbackKey);
+  button.setAttribute('callbackKey', callbackKey);
```
samelhusseini, line 120: You’ll have issues with deserializing this in IE. It doesn’t like cases in XML nodes.

samelhusseini: Oh nevermind, we deal with that case here: https://github.com/google/blockly/blob/23a78c89e4c0f2801768b5c55c7f91ac261f4bc6/core/flyout_button.js#L80

PR #261, opened 2019-10-24
Update gulp publish command, never serialize image field

shakao:
@samelhusseini, do you remember why the field was being set to editable [here](https://github.com/microsoft/pxt-blockly/commit/2a1b39c130a7faeff4ffb0614310e899acfb1548#diff-b56753ee0b760364c5854d360ff51d61R59)?


jwunderl (approved):
LGTM, but maybe worth waiting on if there was a reason to set it that way in the first place

riknoll (approved):

samelhusseini:
Looks like at that time fieldClick was dependent on EDITABLE being true. https://github.com/microsoft/pxt-blockly/blob/2a1b39c130a7faeff4ffb0614310e899acfb1548/core/gesture.js#L749

That is no longer the case. That being said, there's an issue here because what you really want is EDITABLE be true, and SERIALIZABLE be false, but this code [here](https://github.com/google/blockly/blob/master/core/field.js#L434) doesn't let you do that.
I think what you're doing here is fine for now, I'll file a bug on Blockly core and address.

shakao:
@samelhusseini Awesome, thanks Sam!

PR #274, opened 2020-03-25
Custom rendering for collapsed block with statement input

shakao:
- Adds CollapsedInputStack row renderable to the render info if the block contains a statement input, otherwise adds the JaggedEdge as before
- Only displays collapse/expand context menu items for top-level blocks
- Small fix for bubbling warnings up from collapsed blocks

collapse_icon


core/renderers/common/constants.js
```diff
@@ -241,6 +241,18 @@ Blockly.blockRendering.ConstantProvider = function() {
    */
   this.JAGGED_TEETH_WIDTH = 6;
 
+  /**
```
samelhusseini: In order to make it easier to merge things in the future, why not add this in the pxt renderer? 
You can add these properties in the constructor of https://github.com/microsoft/pxt-blockly/blob/develop/core/renderers/pxt/constants.js
And then override init to call super and make the ellipses.

core/renderers/common/constants.js
```diff
@@ -725,6 +742,31 @@ Blockly.blockRendering.ConstantProvider.prototype.makeJaggedTeeth = function() {
   };
 };
 
+/**
+ * pxt-blockly
+ * @return {!string} A string containing path information about
+ *     collapsed block ellipses.
+ * @package
+ */
+Blockly.blockRendering.ConstantProvider.prototype.makeEllipses = function() {
```
samelhusseini: As mentioned above, this would be better fit in the pxt constant provider.

core/renderers/measurables/rows.js
```diff
@@ -541,3 +548,29 @@ Blockly.blockRendering.InputRow.prototype.measure = function() {
...
+/**
+ * pxt-blockly An object containing information about a row representing a
+ *   collapsed statement input.
+ * @param {!Blockly.blockRendering.ConstantProvider} constants The rendering
+ *   constants provider.
+ * @package
+ * @constructor
+ * @extends {Blockly.blockRendering.Row}
+ */
+Blockly.blockRendering.CollapsedInputRow = function(constants) {
```
samelhusseini: I would also move this into a pxt renderer measurable. See how its done in the zelos renderer: 
https://github.com/google/blockly/blob/master/core/renderers/zelos/measurables/row_elements.js

shakao:
Fix for https://github.com/microsoft/pxt-microbit/issues/2672 bundled in, needed for adding the collapse chevron to all function defs

riknoll (approved):

PR #282, opened 2020-04-07
Adding go to definition and removing edit function from calls

### LINKED ISSUE #2771: Go to Definition
**Describe the bug**
When a program is large enough, it is impossible to find the definition of a function. Go to definition in the content menu would go a long way.

The call block has "edit function" which should go replaced by "Go to Function"
image

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2771

This adds a "go to definition" option on the context menu for function calls and removes "edit function" so that now it's only on definitions.


core/pxt_blockly_functions.js
```diff
@@ -395,6 +395,19 @@ Blockly.Functions.makeEditOption = function(block) {
   return editOption;
 };
 
+Blockly.Functions.makeGoToDefinitionOption = function(block) {
+  var gtdOption = {
+    enabled: true,
+    text: Blockly.Msg.FUNCTIONS_GO_TO_DEFINITION_OPTION,
+    callback: function() {
+      var functionName = block.getField("function_name").getText();
```
shakao, line 403: should this be getValue?

riknoll: should it? I just copied what we do in blocklycompiler

shakao: HMM no i think you're right, gettext is safer for compatibility with old functions

shakao (approved):

abchatra:
@tballmsft  for you

PR #287, opened 2020-04-24
Reindeer connection indicator with line

shakao:
feature-reindeer


pelikhan:
game changer!

pelikhan (approved):

samelhusseini:
Brilliant!

jwunderl (approved):
Looks fantastic.

very minor nitpick, the right border on the `+` dropdown showing up on top of the block stands out to me a bit, if it's easy to move the reindeer connector to be a little higher

shakao:
@jwunderl yeah, so the connection indicator is part of each field (in this case the number and not the parent "plus" block) so this is due to how the field elements are added to the block, as svg layering is based on the order of child elements. it's probably possible to reorder the dom based on which field is being highlighted but i feel like the risks of that outweigh the rewards haha

riknoll:
Looks amazing!

yonatanm:
@shakao - L :heart: ve it. super useful!
@samelhusseini - this change is not limited to the renderer only (core/gestures.js and core/insertion_marker_manager.js ), is it possible to get this functionality in Blockly w/o forking it?

PR #294, opened 2020-05-01
Add grayscale filter for block

shakao:
put this together really quick--right now it just applies it to the individual block, i'm not sure how the blocks diff works but you probably traverse the children yourself?

if you do 'gulp publish --closure-library' you should be able to build this branch locally for pxt and test it out

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1324,6 +1324,18 @@ Blockly.BlockSvg.prototype.setHighlighted = function(highlighted) {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


pelikhan (approved):
Yes I can recourse the tree

pelikhan:
It does not seem to apply to dropdowns properly. The setColour works. Can we make a bit lighter?
image

raghavthind2005 (approved):

PR #303, opened 2020-05-21
Put new functions on center of workspace view instead of at the top

### LINKED ISSUE #3111: New function should appear in the middle of the view
We place new functions above on start which might be far away from the current screen. It should dropped in the center of the current view.

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3111

This removes the annoying behavior of always putting function blocks above the "highest" block in the workspace. Now we place it at the center of the current view. One downside to this is that now the function will often be placed on top of other blocks, but it's easy enough to move it.


jwunderl (approved):

shakao:
HOORAY

shakao (approved):

PR #310, opened 2020-05-28
Replace backslashes with forward slashes in generated closure depende…

shakao:
Only an issue for windows builds. Probably should at some point investigate if there's a way to deploy the Travis build artifacts instead of checking in the built files


jwunderl (approved):

samelhusseini:
Is this for uncompressed?
Recent change for context on the build side: https://github.com/google/blockly/pull/3894

livcheerful (approved):

shakao:
yep, just for uncompressed--and cool! we should be able to port all your gulp updates when we next merge so hopefully this change can be reverted after haha

PR #315, opened 2020-07-09
domToText: don't run regexp over whole document

eanders-ms:
Blockly's domToText method came up as a hot path due to a regexp replace it executes iteratively over the entire xml. This regexp was added when multline text blocks came in in order to flatten the plain-text representation of the serialized dom (this is a limitation of the plaintext format).

Testing with this game: https://arcade.makecode.com/74548-67333-76127-33508

Time spent in `saveWorkspaceXml`:
| Before | After |
|---|---|
| 10s | 250ms |


samelhusseini:
Good find! Would you mind PR'ing against blockly core?

darzu (approved):

shakao (approved):

eanders-ms:
@samelhusseini Will do!

core/field_multilineinput.js
```diff
@@ -91,6 +91,21 @@ Blockly.FieldMultilineInput.fromJson = function(options) {
...
+ * @param {!Element} fieldElement The element to populate with info about the
+ *    field's state.
+ * @return {!Element} The element containing info about the field's state.
+ * @package
+ */
+Blockly.FieldMultilineInput.prototype.toXml = function(fieldElement) {
+  // Replace '\n' characters with html-escaped equivalent '&#10'. This is needed
+  // so the plain-text representation of the xml produced by `Blockly.Xml.domToText`
+  // will appear on a single line (this is a limitation of the plain-text format).
+  fieldElement.textContent = this.getValue().replace(/\n/g, '&#10');
```
eanders-ms, line 105: Oh this isn't quite right. Needs to be `&#10;`

mattbishop: Shouldn't this use the original regex?

samelhusseini: I don't think it needs to, because the original regex was looking over all of the XML, whereas this one only gets the multi-line field's value.

eanders-ms: The original regexp was matching xml open/close tags. At this scope we're just operating on the content string (what appears between the xml tags).

mattbishop: All right, and as a suggestion, what about wrapping the content in `<![CDATA[ ... ]]>` instead? Might be faster too.

eanders-ms: This change brings the performance of domToText to an acceptable level for MakeCode. We could try the CDATA approach if additional perf issues arise in serialization. I admit my understanding of when to use CDATA is spotty. I think it is useful when you want to embed XML in a content string.

core/xml.js
```diff
@@ -305,17 +305,6 @@ Blockly.Xml.cloneShadow_ = function(shadow, opt_noId) {
  */
 Blockly.Xml.domToText = function(dom) {
   var text = Blockly.utils.xml.domToText(dom);
```
BeksOmega, line 307: I think you could make this just `return Blockly.utils.xml.domToText(dom);` =)

PR #317, opened 2020-07-14
Add fromXml override to FieldMultilineInput

eanders-ms:
Missed this in my previous commit. I was testing in playground.html and noticed that I had to decode the content string in the corresponding `fromXml` override.

Related PRs: #315, #316 

Also reformatted comments to wrap at col 80 (as per Blockly's style guide).


eanders-ms:
@abchatra If my previous commits (mentioned above) are going out in the next release, this one should also be included.

abchatra:
Just bump the Blockly npm dependency in pxt. It will be included in next release

abchatra:
At package.json in pxt

PR #320, opened 2020-08-21
Don't allow rename to blank string

shakao:
The function rename code was updated to support function duplication and missed this case! If the name passed in is an empty string (at some point we were storing the function name on the mutation and not in the function name element), we return the old name.

Fixes https://github.com/microsoft/pxt-arcade/issues/2351


jwunderl (approved):

riknoll:
can you still delete all the text while typing? like if I wanted to completely remove the old string and type a new one would it force me to have at least one character?

shakao:
yeah, it doesn't run the rename code until the field loses focus (same for renaming, you can have "doSomething" and type "doSomething" and on enter/clicking out it renames to "doSomething2)--if you hit enter on an empty string it'll just assign the previous name

riknoll (approved):

PR #322, opened 2020-09-03
Add "Create 'call <function>'" to function context menu.

jacobcarpenter:
Inspired by @riknoll's awesome microsoft/pxt#7445, I decided to take a crack at microsoft/pxt-arcade#2094. 😊 

I've only tested this in the `tests/playground.html`, but wanted to open the discussion on whether this is useful/desirable/the right place to approach adding this feature. (It seems like a nice to have feature to me!)

No worries if you don't want it or it's not the right approach or anything. Feedback requested @shakao @riknoll. Thanks!

Kapture 2020-09-03 at 11 20 37

*Edit:* added gif.

### DIFF
--- a/blocks/pxt_blockly_functions.js
+++ b/blocks/pxt_blockly_functions.js
@@ -997,7 +997,6 @@ Blockly.Blocks['function_definition'] = {
    */
   init: function() {
     this.jsonInit({
-      "extensions": ["function_contextmenu_edit"],
       "style": {
         "hat": "cap"
       }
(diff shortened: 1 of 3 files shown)


riknoll (approved):
LGTM! Thanks as always!

shakao (approved):
this is awesome!! (also sorry about the build being broken haha, if you rebase/merge now it should be fixed!)

jacobcarpenter:
> if you rebase/merge now it should be fixed!

Indeed!

Will localization strings automatically be scheduled for translation or is there something I'd need to do for that?

Is there any corresponding work I'd need to do in `microsoft/pxt` if this gets merged, or is that automatic (or something you'd take care of)?

riknoll:
Yup, we'll take care of it when we do a release of pxt-blockly. The changes on the PXT side are that we need to add it to our localization system. Here are the relevant bits of code:

1. Add the string to our localization here: https://github.com/microsoft/pxt/blob/master/pxtlib/blocks.ts#L702
2. Patch the msg here: https://github.com/microsoft/pxt/blob/master/pxtblocks/blocklyloader.ts#L2424

And that should take care of it.

PR #325, opened 2020-09-03
Enable go to definition for the context menu of function calls

riknoll:
Classic "forgetting to call the function" moment.

Fixes https://github.com/microsoft/pxt-arcade/issues/2412


jwunderl (approved):

shakao:
could just delete that "Edit" option entirely?

riknoll:
Do we want to do that? I certainly can

shakao:
yeah imo people rarely use the edit anyway? plus you can easily go to def -> edit, even if it's one extra click

riknoll:
will do

livcheerful (approved):

shakao (approved):

PR #328, opened 2020-11-12
Show modal for text input with touch

### LINKED ISSUE #3496: On Screen Keyboard issues

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3496

The issue was we were first focusing on the workspace, then the text input, and then the workspace again, but instead of looking more into that, I saw that we were using a prompt editor for mobile browsers because of focus and on screen keyboard issues, so I thought making general touch events trigger it would be a good fix. If you guys think it's better to keep it inline then I can do it the other way :D 

(this is what it looks like on my Surface)
image

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -386,14 +386,11 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(_opt_e,
   this.workspace_ = this.sourceBlock_.workspace;
   var quietInput = opt_quietInput || false;
   var readOnly = opt_readOnly || false;
-  var isTouchEvent = undefined;
-  if (_opt_e instanceof PointerEvent) {
-    isTouchEvent = _opt_e.pointerType === "touch";
-  }
+  var isTouchEvent = Blockly.Touch.isTouchEvent(_opt_e);
 
   if (!quietInput && (Blockly.utils.userAgent.MOBILE ||
                       Blockly.utils.userAgent.ANDROID ||
-                      Blockly.utils.userAgent.IPAD||
+                      Blockly.utils.userAgent.IPAD ||
                       isTouchEvent)) {
     this.showPromptEditor_();
   } else {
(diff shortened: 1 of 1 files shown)


core/field_textinput.js
```diff
@@ -386,9 +386,15 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(_opt_e,
...
   var readOnly = opt_readOnly || false;
+  var isTouchEvent = undefined;
+  if (_opt_e instanceof PointerEvent) {
+    isTouchEvent = _opt_e.pointerType === "touch";
+  }
+
   if (!quietInput && (Blockly.utils.userAgent.MOBILE ||
                       Blockly.utils.userAgent.ANDROID ||
-                      Blockly.utils.userAgent.IPAD)) {
+                      Blockly.utils.userAgent.IPAD) ||
```
jwunderl: I think the `|| isTouchEvent` should be with the other ios events inside the paranetheses, so quietInput has an effect?

jwunderl (approved):

core/field_textinput.js
livcheerful: oh shoot yes something got messed up ty joey

shakao:
blockly has a builtin "isTouchEvent" function: https://github.com/microsoft/pxt-blockly/blob/develop/core/touch.js#L236 but does this preserve the current drag behavior? right now i can drag a block even if the first place i "touched" is a text input, would this pop up the keyboard instead?

livcheerful:
> blockly has a builtin "isTouchEvent" function: https://github.com/microsoft/pxt-blockly/blob/develop/core/touch.js#L236 but does this preserve the current drag behavior? right now i can drag a block even if the first place i "touched" is a text input, would this pop up the keyboard instead?

This is excellent :O And I just tested it with microbit -- drag events still work even if I'm dragging by the text field XD it's only very precise, no planar movement that will trigger this.

shakao (approved):

PR #331, opened 2020-12-17
Adding workspace loaded method to fields

riknoll:
This adds an "onWorkspaceLoaded" method for fields that has the following semantics:

1. If the workspace is being loaded from XML, this will fire for all fields after all blocks are initialized
2. If the workspace is already loaded, this will fire after all fields in the block have been initialized

The reasoning for this is that it's pretty difficult right now to know when a field has received the "real" value while it's being created.

When Blockly creates blocks (e.g. when loading a workspace from XML), it often calls setValue multiple times with different values. This causes a lot of problems in the field editors in PXT that are tied to resources outside of the Blockly workspace. For example, we have image editors for which the source of truth is a file in the user's project. We need to know when we have the real value so that we can be sure it's safe to change those files based on the value in the field editor.

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -48,6 +48,11 @@ Blockly.Workspace = function(opt_options) {
   /** @type {number} */
   this.toolboxPosition = this.options.toolboxPosition;
 
+  // pxtblockly: used to prevent workspace load notifcations from happening while
+  // workspace is still being loaded from xml
+  /** @type {boolean} */
+  this.loadingEventsDisabled = false;
+
   /**
    * @type {!Array.<!Blockly.Block>}
    * @private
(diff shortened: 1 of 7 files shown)


core/workspace.js
```diff
@@ -48,6 +48,11 @@ Blockly.Workspace = function(opt_options) {
   /** @type {number} */
   this.toolboxPosition = this.options.toolboxPosition;
 
+  // pxtblockly: used to prevent workspace load notifcations from happening while
```
jwunderl:
```suggestion
  // pxtblockly: used to prevent workspace load notifications from happening while
```

nit

jwunderl (approved):

livcheerful (approved):

PR #332, opened 2020-12-18
Check pointer events to see if they are triggered by a mouse

### LINKED ISSUE #3496: On Screen Keyboard issues

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3496

The way Blockly was detecting touch events (by seeing if it was a PointerEvent or a TouchEvent and not a MouseEvent) doesn't work anymore since Chrome will send PointerEvents always. 

This change will see what is the source of the PointerEvent (mouse, pen, or touch) to decide if it is a touch event or not


livcheerful:
Opening a new one with build errors fixed

PR #333, opened 2020-12-18
Check pointer events to see if they are triggered by a mouse 

### LINKED ISSUE #3496: On Screen Keyboard issues

livcheerful:
fixes microsoft/pxt-microbit#3496

The way Blockly was detecting touch events (by seeing if it was a PointerEvent or a TouchEvent and not a MouseEvent) doesn't work anymore since Chrome will send PointerEvents always.

This change will see what is the source of the PointerEvent (mouse, pen, or touch) to decide if it is a touch event or not

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -386,7 +386,9 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(_opt_e,
   this.workspace_ = this.sourceBlock_.workspace;
   var quietInput = opt_quietInput || false;
   var readOnly = opt_readOnly || false;
-  var isTouchEvent = Blockly.Touch.isTouchEvent(_opt_e);
+  var isTouchEvent = false;
+  if (_opt_e)
+    isTouchEvent = Blockly.Touch.isTouchEvent(_opt_e);
 
   if (!quietInput && (Blockly.utils.userAgent.MOBILE ||
                       Blockly.utils.userAgent.ANDROID ||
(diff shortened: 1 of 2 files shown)


jwunderl (commented):
Could you do an uploadtrg of arcade with this change so we can test on different devices? Looks fine to me but click / touch event things like this feel like they always end up needing special casing for at least one browser

livcheerful:
> Could you do an uploadtrg of arcade with this change so we can test on different devices? Looks fine to me but click / touch event things like this feel like they always end up needing special casing for at least one browser

ohhh for sure for sure I should have thought of that!

livcheerful:
@jwunderl 🤲 i have an uploadtrg!

https://arcade.makecode.com/app/5d3b8eda3593ee3716ff67d11caea43c60804482-51ba2d86b6#editor

i'm gonna look into the position block and see if i can pass in an event!

jwunderl (approved):
Tested on browerstack and it seems good to me! If you have a physical ios device would be good to double check on that as well.

While testing I did notice we have a focus issue with the blockly text entry dialog on mobile in a few of the field editors (e.g. if you open the position picker or `set velocity to vx vy` it will pop the dialog but lose focus on the text, so you have to click again to refocus) but that's a separate issue, will file after you merge this~

livcheerful:
> Tested on browerstack and it seems good to me! If you have a physical ios device would be good to double check on that as well.
> 
> While testing I did notice we have a focus issue with the blockly text entry dialog on mobile in a few of the field editors (e.g. if you open the position picker or `set velocity to vx vy` it will pop the dialog but lose focus on the text, so you have to click again to refocus) but that's a separate issue, will file after you merge this~

thanks joey! i will test on my ipad!

livcheerful:
I tried with Chrome & Safari on my iPad and it looks alright! I'm gonna complete this PR

PR #336, opened 2021-01-13
Don't focus on workspace every time blockly is resized

livcheerful:
this is the freshest one try this one https://arcade.makecode.com/app/56ea276e327b4140571532c6e0a42035c2e37f12-31bcb065cf

(affects https://github.com/microsoft/pxt-arcade/issues/2792 
https://github.com/microsoft/pxt-arcade/issues/2793
https://github.com/microsoft/pxt-arcade/issues/2794
also https://github.com/microsoft/pxt-arcade/issues/2434 probably i gotta test)

This fix is to check if the resize is from the on screen keyboard popping up, and if it is, not hiding WidgetDiv and DropDownDiv since they will automatically focus on the workspace after hiding.

I also got rid of the modal change from before. The new behavior is:
Clicking with a mouse: is always inline
Clicking with touch on desktop: is always inline, on screen keyboard shows
Clicking with touch on mobile: modal

New uploadtrg!
https://arcade.makecode.com/app/56ea276e327b4140571532c6e0a42035c2e37f12-31bcb065cf

AND Here's a project with a bunch of inputs to test with https://makecode.com/_fPcPbXfs3F2F

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -588,7 +588,8 @@ Blockly.utils.getBlockTypeCounts = function(block, opt_stripFollowing) {
 };
 
 var resizeFromKeyboard = true;
-var initialWidth = 0;
+var initialWidth = 0; // The width of the window before resizing started
+var lastHeight = 0; // The height at each frame of the resize
 var debounceTimeout = null;
 
 /**
@@ -601,15 +602,35 @@ Blockly.utils.resizeTracker = function(workspace) {
     clearTimeout(debounceTimeout);
   }
   debounceTimeout = setTimeout(function(){
+    var injectionDiv = workspace.getInjectionDiv();
     resizeFromKeyboard = true;
-    initialWidth = workspace.width;
+    initialWidth = injectionDiv.clientWidth;
+    lastHeight = injectionDiv.clientHeight;
   }, 1000);
 
-  if (resizeFromKeyboard) {
-    // If the on-screen keyboard is causing the resize, only the
-    // height will change.
-    if (workspace.width != initialWidth) {
-      resizeFromKeyboard = false;
+  var injectionDiv = workspace.getInjectionDiv();
+  if (injectionDiv) {
+    if (initialWidth == 0) {
+      // Initialize the correct values
+      resizeFromKeyboard = true;
+      initialWidth = injectionDiv.clientWidth;
+      lastHeight = injectionDiv.clientHeight;
+    }
+
+    if (resizeFromKeyboard) {
+      // If the on-screen keyboard is causing the resize, only the
+      // height will change.
+      var currWidth = injectionDiv.clientWidth;
+      var currHeight = injectionDiv.clientHeight;
+
+      if (currWidth != initialWidth) {
+        resizeFromKeyboard = false;
+      }
+      if (currHeight > lastHeight) {
+        resizeFromKeyboard = false;
+      } else {
+        lastHeight = currHeight;
+      }
     }
   }
 }
@@ -619,7 +640,7 @@ Blockly.utils.resizeTracker = function(workspace) {
  * is from an on-screen keyboard appearing.
  */
 Blockly.utils.isOnScreenKeyboardResize = function() {
-  if (document.activeElement.className == "blocklyHtmlInput" && resizeFromKeyboard) {
+  if (Blockly.utils.dom.hasClass(document.activeElement, "blocklyHtmlInput") && resizeFromKeyboard) {
     // If it is an input element, the resize might be triggered by the
     // onscreen keyboard, but we also need to make sure only the height changed
     return true;
--- a/core/utils.js
+++ b/core/utils.js
@@ -587,6 +587,47 @@ Blockly.utils.getBlockTypeCounts = function(block, opt_stripFollowing) {
   return typeCountsMap;
 };
 
+var resizeFromKeyboard = true;
+var initialWidth = 0;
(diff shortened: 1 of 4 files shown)


core/utils.js
```diff
@@ -587,6 +587,47 @@ Blockly.utils.getBlockTypeCounts = function(block, opt_stripFollowing) {
...
+  }
+  debounceTimeout = setTimeout(function(){
+    resizeFromKeyboard = true;
+    initialWidth = workspace.width;
+  }, 1000);
+
+  if (resizeFromKeyboard) {
+    // If the on-screen keyboard is causing the resize, only the
+    // height will change.
+    if (workspace.width != initialWidth) {
```
shakao: won't this trigger if the user resizes the browser window vertically (with a mouse) as well, though?

core/utils.js
```diff
@@ -587,6 +587,47 @@ Blockly.utils.getBlockTypeCounts = function(block, opt_stripFollowing) {
...
+    }
+  }
+}
+
+/**
+ * Use during resize to see if the source of the resize
+ * is from an on-screen keyboard appearing.
+ */
+Blockly.utils.isOnScreenKeyboardResize = function() {
+  if (document.activeElement.className == "blocklyHtmlInput" && resizeFromKeyboard) {
```
shakao: should probably use "hasClass" https://github.com/microsoft/pxt-blockly/blob/develop/core/utils/dom.js#L164 since the className is a string of all classes on the element & could change in the future

core/utils.js
livcheerful: ah yeah that's true. Hmmm maybe I could look at how fast it is? Since the keyboard probably changes it fast and in only one direction

shakao: is this function https://github.com/microsoft/pxt-blockly/blob/develop/core/dropdowndiv.js#L730 not successfully preserving the visibility of the widget/dropdown divs?

livcheerful: OO I didn't know about that function. It's not working I think because it's being hidden first (but I gotta check that just going off my gut)

shakao:
here's another PR joey did related to resizing the workspace/on-screen keyboard, not sure if it helps? https://github.com/microsoft/pxt-blockly/pull/305

livcheerful:
:O thanks shannon! that's the same issue we're having now. I'll see why that doesn't work anymore...

livcheerful:
ah i does work still! just for non-blockly fields like editing the project name hmmmmmmmmmmmmmmmm

core/utils.js
livcheerful: (oh yes i looked into this and even if we did reposition the dropdown correctly after, it doesn't matter since we've already refocused onto the workspace, and if we tried to focus back onto the input we'd start the on screen keyboard popping up/down infinitely)

and for the original comment, I think we're actually okay from testing it out. if the window gets too small, the dropdown gets closed, and if we're using touch, the onscreen keyboard will actually block the bottom edge so it can't be used to resize

core/utils.js
livcheerful: fixed! thanks shannon!

core/utils.js
shakao: hmm maybe i'm missing something? if i resize the workspace up horizontally won't that permanently set "resizeFromKeyboard" to true?

shakao: hmm ok but i guess, worst case is the widgets will remain onscreen and not be hidden, which is probably not terrible

livcheerful: permanently for that resize, not permanently forever (not sure if that was what you mean?)

since the only time we're checking to see if it's resizing from the keyboard is when we're deciding to hide the widget/dropdown, the worst that could happen is it wouldn't properly hide it when it gets too small to fit -- but for some reason when I was testing it, the dropdown was still automatically closed when it wouldnt fit anymore

vertical resize

not sure why it works though???

or I might not be understanding your concern correctly >>

core/utils.js
```diff
@@ -587,6 +587,68 @@ Blockly.utils.getBlockTypeCounts = function(block, opt_stripFollowing) {
...
+ * @param {!Blockly.WorkspaceSvg} workspace Any workspace in the SVG.
+ */
+Blockly.utils.resizeTracker = function(workspace) {
+  if (debounceTimeout) {
+    clearTimeout(debounceTimeout);
+  }
+  debounceTimeout = setTimeout(function(){
+    var injectionDiv = workspace.getInjectionDiv();
+    resizeFromKeyboard = true;
+    initialWidth = injectionDiv.clientWidth;
```
shakao: does this need to also be in an `if (injectionDiv)` check?

shakao:
one small comment otherwise lgtm

shakao (approved):

core/utils.js
shakao: the one other case we have for vertical resizing is the more/less for the tutorial step, so if the tutorial expands the widget might cover the tutorial card, but that should be pretty easy to recover from also

livcheerful: oooo i see i'll try it out though!

livcheerful: hmmm yeah so it doesn't close with the tutorials, but I just tried it out on live too and that doesn't work either :\

shakao: not a regression! seems fine to me then :D

core/utils.js
livcheerful: fixed!

jwunderl (approved):

PR #339, opened 2021-03-06
Fix mobile slider bugs

jwunderl:
Fix a few issues with mobile sliders at the moment, I'll annotate them inline as they're all fairly minor / subtle. Here's a minecraft build with these changes https://minecraft.makecode.com/app/addcfbcdcff844377d13ddbf137722a04871e317-b174e696c4


core/css.js
```diff
@@ -917,7 +917,7 @@ Blockly.Css.CONTENT = [
   '.blocklyDropDownDiv .goog-slider-horizontal .goog-slider-thumb {',
      'width: 26px;',
      'height: 26px;',
-     'margin-top: -1px;',
+     'top: -1px;',
```
jwunderl, line 920: This is to work around a bug in the closure library version being used, but is effectively identical behavior wise (breaking the link with a space so it doesn't do the annoying backref thing thing https://github.com/google/closure-library /pull/877)

core/field_slider.js
```diff
@@ -138,7 +138,7 @@ Blockly.FieldSlider.prototype.init = function() {
  * @private
  */
 Blockly.FieldSlider.prototype.showEditor_ = function(e) {
-  Blockly.FieldSlider.superClass_.showEditor_.call(this, e, false);
+  Blockly.FieldSlider.superClass_.showEditor_.call(this, e);
```
jwunderl, line 141: Passing false here means all the sliders pop out to text inputs here https://github.com/microsoft/pxt-blockly/blob/ad5295553d2d0369b0d30160f84e42ad19670901/core/field_number.js#L339-L346, which is not great. leaving it as default lets it use the standard mobile regexes to check if it should be silent or not.

Technically this means it will hit the 'showNumpad' code, but it never actually shows up as the slider itself clears it. If it's a concern I can add a second parameter for `don't show numpad, but also be quiet on mobile` if we want.

core/field_slider.js
```diff
@@ -180,10 +180,10 @@ Blockly.FieldSlider.prototype.addSlider_ = function(contentDiv) {
     this.readout_ = elements[1];
   }
   this.slider_ = new goog.ui.Slider();
-  this.slider_.setMoveToPointEnabled(false);
   this.slider_.setMinimum(this.min_);
   this.slider_.setMaximum(this.max_);
-  if (this.step_) this.slider_.setUnitIncrement(this.step_);
+  this.slider_.setMoveToPointEnabled(!this.step_);
```
jwunderl, line 185: setMoveToPoint makes it so it snaps to the position, so this would make it so it snaps directly to where you click if you click on the bar (which feels a lot better than 'move towards where you clicked in 10 unit increments' to me). Can revert or make it optional if we're not sure on that.

core/field_slider.js
```diff
@@ -180,10 +180,10 @@ Blockly.FieldSlider.prototype.addSlider_ = function(contentDiv) {
...
+  if (this.step_) this.slider_.setBlockIncrement(this.step_);
```
jwunderl, line 186: re: https://google.github.io/closure-library/api/goog.ui.Slider.html; unitIncrement is used but doesn't have a visible change in this case, it's for arrow keys / scroll wheel / etc. The value that is meaningful for us is 'blockIncrement' which is what's used for large jumps, including the 'move towards where you clicked by n amount'

shakao (approved):

shakao:
lgtm! the blockly build unfortunately hasn't been migrated to github actions yet though

livcheerful (approved):

jwunderl:
I bumped the chromedriver version as well to make travis happy again~

PR #341, opened 2021-03-16
Remove edit options from context menus when in debug mode

riknoll:
Part of the fix for https://github.com/microsoft/pxt-arcade/issues/3220

Removes all options that could be used to edit the workspace when in debug mode. That includes:

1. Delete block(s)
2. Create get/set "var"
3. Edit function
4. Create call "function"
5. Rename variable
6. The list of variable names you get in some of the draggable variables in loops

### DIFF
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -259,13 +259,13 @@ Blockly.Constants.Loops.CUSTOM_CONTEXT_MENU_CREATE_VARIABLES_GET_MIXIN = {
    * @this {Blockly.Block}
    */
   customContextMenu: function(options) {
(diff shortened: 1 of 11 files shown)


jwunderl (approved):
Woops, I hadn't gotten around to getting a new npm token for pxt-blockly yet so will need to do that to get a release for this uploaded. I can do it tomorrow~

riknoll:
@jwunderl please re-review this. I changed a lot of stuff so that now instead removing the option altogether, I just disable it.

jwunderl (approved):

PR #348, opened 2021-05-14
Add arrays to function parameters

livcheerful:
part of https://github.com/microsoft/pxt-microbit/issues/2205

pxt PR: https://github.com/microsoft/pxt/pull/8156

same pictures:
image
image
image
image

shadow block is variable picker
image

### DIFF
--- a/blocks/pxt_blockly_functions.js
+++ b/blocks/pxt_blockly_functions.js
@@ -463,6 +463,11 @@ Blockly.PXTBlockly.FunctionUtils.getShadowBlockInfoFromType_ = function(argument
       fieldName = 'TEXT';
       fieldValue = 'abc';
       break;
+    case 'Array':
+      shadowType = 'variables_get';
+      fieldName = "VAR";
+      fieldValue = Blockly.Variables.getOrCreateVariablePackage(ws, null, "list", '').getId();
+      break;
     default:
       // This is probably a custom type. Use a variable as the shadow.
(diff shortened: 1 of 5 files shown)


jwunderl (approved):
LGTM but probably good to wait for Shannon to approve too ~

blocks/pxt_blockly_functions.js
```diff
@@ -463,6 +463,11 @@ Blockly.PXTBlockly.FunctionUtils.getShadowBlockInfoFromType_ = function(argument
       fieldName = 'TEXT';
       fieldValue = 'abc';
       break;
+    case 'Array':
+      shadowType = 'variables_get';
+      fieldName = "VAR";
+      fieldValue = Blockly.Variables.getOrCreateVariablePackage(ws, null, "list", '').getId();
```
shakao: is there a reason for this to be it's own case instead of an entry in the argumentDefaultNames map?

shakao: (small) and should it use `Blockly.Msg.FUNCTIONS_DEFAULT_ARRAY_ARG_NAME` ?

shakao (approved):

blocks/pxt_blockly_functions.js
livcheerful: oh I just thought since it was going to be one of our default types instead of a custom type, it'd be nice to break it out, but I can be convinced either way!

PR #349, opened 2021-05-17
Use "list" default name

livcheerful:
Fix for https://github.com/microsoft/pxt-blockly/pull/348

Using Blockly.Msg.FUNCTIONS_DEFAULT_ARRAY_ARG_NAME didn't work -- it would just make an 'i' variable, so using "list" instead. I also tried out defining Arrays as part of "getArgumentDefaultName" using pxtarget in microbit, but then it doesn't play nicely with Array being a default type.

I can remove Arrays as defaults and make them defined target by target if that's preferred instead!


shakao (approved):

shakao:
no worries, this looks good to me!

PR #351, opened 2021-06-17
Array block fixes for datalogger

jwunderl:
* Make arrays switch to vertical layout after 4+ slots have been added by default (vs. 6 before) https://github.com/microsoft/pxt-microbit/issues/4177
* Makes that configurable as fix for https://github.com/microsoft/pxt-microbit/issues/4177 (the builds below have `mut.setAttribute("horizontalafter", "" + itemCount);` set here: https://github.com/microsoft/pxt/blob/30a0beeea9e36a6bdb2fa5cd065380d036446d10/pxtblocks/blocklyloader.ts#L271 for example to make it so  non-strings / boolean shadows switch to vertical layout after 3+ instead). If we want it to always go to vertical for that block I'll just need to add a blockattribute for that I suppose
* Fill out nested shadow values one layer deep (shouldn't be any harder to make it recursive, but that feels a little unnecessary / I don't think it hits anything currently anyways?) -- fix for https://github.com/microsoft/pxt-microbit/issues/4161
* If the first slot in the array doesn't have a shadow or has a shadow that doesn't match the current value (e.g. it's a number and value is an array), attach a shallow clone of the visible block as a field -- fix for https://github.com/microsoft/pxt-arcade/issues/1779, but it's very easy to move to a separate pr if it needs discussion.

microbit build: https://makecode.microbit.org/app/2393d789a615db76bdf3d3d53e56acec72ba9685-5df27094a9
arcade build: https://arcade.makecode.com/app/4247fc12189539fbfc2a88a44535b149524b397f-91a6e67c53

### DIFF
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -265,16 +265,25 @@ Blockly.Blocks['lists_create_with'] = {
     this.restoreConnections_();
     // Add shadow block
     if (this.itemCount_ > 1) {
+
       // Find shadow type
       var firstInput = this.getInput('ADD' + 0);
       if (firstInput && firstInput.connection.targetConnection) {
         // Create a new shadow DOM with the same type as the first input
         // but with an empty default value
         var newInput = this.getInput('ADD' + (this.itemCount_ - 1));
-        var shadowInputDom = firstInput.connection.getShadowDom();
-        if (shadowInputDom) {
+        var connectedBlock = firstInput.connection.targetConnection;
+        var connectedBlockType = connectedBlock.getSourceBlock();
+
+        var newShadowType = connectedBlockType && connectedBlockType.type;
+        if (!newShadowType) {
+          var shadowInputDom = firstInput.connection.getShadowDom();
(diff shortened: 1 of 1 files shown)


blocks/lists.js
```diff
@@ -264,20 +272,72 @@ Blockly.Blocks['lists_create_with'] = {
...
-            newInput.connection.respawnShadow_();
+        var newShadowType = shadowInputDom && shadowInputDom.getAttribute('type');
+        if (newShadowType) {
+          var shadowDom = createBlockDom('shadow', newShadowType);
+          var childValues = getChildrenByTag(shadowInputDom, 'value');
+
+          for (var i = 0; i < childValues.length; i++) {
+            var input = childValues[i];
+            var inputName = input.getAttribute('name');
+            var fieldShadow = getChildrenByTag(input, 'shadow')[0];
```
riknoll: should probably call this "valueShadow" (same below) to match the blockly terminology

riknoll (approved):

PR #354, opened 2021-06-30
Check connection before reconnecting function args

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3623

If you recreate an argument input with the same name, verify that whatever was connected before *can* connect to the new input. If it can't, then we turn it into a full block if it's a shadow to prevent it from becoming a top-level shadow block.


eanders-ms (approved):
LGTM

PR #367, opened 2021-11-16
Add Github action for publshing to NPM

shakao:
Fixes the pxt bump gulp taks (`num run pxt:bump`) and adds a github action which runs on tagged commits to publish to npm. I'll make a separate PR manually bumping to a clean version (accidentally published 4.0.1 while testing) then do a bump!

### DIFF
--- a/scripts/gulpfiles/pxt_tasks.js
+++ b/scripts/gulpfiles/pxt_tasks.js
@@ -48,11 +48,15 @@ const pxtTest = gulp.series([buildTasks.core, pxtPublishTask]);
 // Task for bumping patch version and tagging commit. Travis will upload to npm.
 const pxtBump = gulp.series(
   // Sync to latest
-  execSync('git checkout develop', { stdio: 'inherit' }),
-  execSync('git pull origin develop', { stdio: 'inherit' }),
+  function(done) {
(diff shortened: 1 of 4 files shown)


scripts/gulpfiles/pxt_tasks.js
```diff
@@ -42,36 +46,27 @@ function pxtPublishTsTask() {
...
-      .pipe(gulp.git.commit(v))
+      .pipe(gulp.dest('./'))
       .on('end', function () {
-        gulp.git.tag('v' + v, v, function (error) {
-          if (error) {
-            return done(error);
-          }
-          gulp.git.push('origin', '', { args: '--tags' }, done);
-        })
+        execSync('git add .', { stdio: 'inherit' });
```
riknoll: this seems a little dangerous? I feel like bump should be an empty commit

shakao: hmm so it bumps the version number in the package.json and rebuilds the compressed files and typings, but i could move that to be a separate action... or add a specific named set of files?

riknoll: yeah, i think a list would be good

livcheerful (approved):

jwunderl (approved):

PR #381, opened 2022-01-31
Fix function calls population in array

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/3723, it's the issue that pops up when you expand an array where the first element is a user function call and it creates a broken function https://makecode.com/_WR39VPiMhDVh:

brokenfuncadd

and with this:

fixedfuncadd

I am hitting a weird bug, have to check my closure version or something maybe as it repros locally but not in beta / is not directly from this code path; when I right click -> `create 'call doSomething'` twice in a row it replaces the parameter names with the other blocks ids. It does the same when pressing the plus, but we don't hit the list path in the right click menus operation so it... shouldn't repro in live?

ezgif com-gif-maker

 now also fixes https://github.com/microsoft/pxt-arcade/issues/4558

* build error seemed to be spurious / possibly just from dependencies & unrelated to this change, so we disabled it for now; issue here: https://github.com/microsoft/pxt-arcade/issues/4559 tracking turning them on again for whenever we need to make more significant changes~


riknoll (approved):

jwunderl:
Ah hmm that bug I mentioned at the end is actually hitting in beta, not sure why I didn't get it before; it looks like a regression from the blockly update maybe? I'll take a look but @shakao do you possibly remember anything here that could cause this?

shakao:
hmm so there was this change https://github.com/microsoft/pxt-blockly/commit/49ab07ddb6476646672857795876148d4f6bcf8a#diff-b091746f6d9911d9db3d1cf2ea1b3c8a4be77107a9833be64451d4cc3760c039R271 and disconnectOldBlocks is called in updateDisplay which might get hit when you're creating the call block? the other thing is that the entire domToBlock function changed, so if we're creating dom elements and converting those to a block, there might be something different in there

jwunderl:
Ahh figured it out (well, got a fix) -- it looks like https://github.com/microsoft/pxt-blockly/blob/242ff041250940f8d266304e0ac78ed37beca440/blocks/pxt_blockly_functions.js#L630 was always weird for variable getters (that is, it always produced an invalid field), but for some reason it wasn't used before -- that is, the connect just above connected the original variable & so the broken one with no id / just the id as textcontent was left unused. Feels like some deep blockly lifecycle lore that I don't have the will to investigate :)

shakao (approved):

livcheerful (approved):

jwunderl:
hm do we care about the broken build? It's happening in the auto generated pr too, not sure where it would have come from (dependency breaking change / chrome version bump? latter feels unlikely as it repros locally i guess)
image

PR #399, opened 2022-06-07
Only use output shape for output connections

### LINKED ISSUE #4130: Empty field of "is sprite touching other" block looks like boolean
**Describe the bug**
The empty field of "is %sprite|touching %other" block (%other) looks like boolean because the field is hexagon. This is confusing for beginner users.

**To Reproduce**
1. Put "is sprite touching other" block in the programing area

**Expected behavior**
The empty field should be rounded shape.

**Screenshots**
image

**micro:bit version (please complete the following information):**

V2.0

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 91

aznhassan:
#### Problem
The shape of input connections should be decided based on the check type and not the output shape of the source block. This leads to blocks that return a Boolean having all inputs shaped as hexagons as well, even if the inputs are not Booleans themselves.

#### Solution
Only use the output shape of the source block for OUTPUT_VALUE connection types. Use the check type otherwise.

#### Validation
 Built and ran locally using pxt-microbit. The `game_sprite_touching_sprite` block now has circular connections for its sprite inputs vs the hexagonal ones it used to have.

Build: https://microbit.staging.pxt.io/app/d6f9b2143795af11f5046326f410a0f5a9dc4598-667bbfa881#editor

Fixes microsoft/pxt-microbit/issues/4130

### DIFF
--- a/core/renderers/zelos/constants.js
+++ b/core/renderers/zelos/constants.js
@@ -652,7 +652,6 @@ Blockly.zelos.ConstantProvider.prototype.shapeFor = function(
     checks = connection.targetConnection.getCheck();
   }
   switch (connection.type) {
-    case Blockly.connectionTypes.INPUT_VALUE:
     case Blockly.connectionTypes.OUTPUT_VALUE:
       var outputShape = connection.getSourceBlock().getOutputShape();
       // If the block has an output shape set, use that instead.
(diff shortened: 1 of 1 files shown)


riknoll (approved):

jwunderl (approved):

abchatra:
@rachel-fenichel should we fix this in the upstream blockly as well?

rachel-fenichel:
Yes! Thanks.

abchatra:
@aznhassan please produce a PR on the mainline blockly with these changes if you can.

